Issue #8, opened 2019-02-13
Some comments on usage


stiegerb:
Hi @vkuznet,
I'm trying to set up cms-htcondor-es to use the validation code via CMSMonitoring/StompAMQ, and I have some comments on the usage. I'll also prepare a PR with my suggested changes:

- Maybe it would be easier if the validation happened in `make_notification` rather than in the send method. That way one could choose to reject individual documents that fail validation.

- I think the users should be able to specify which schema will be used for validation when instantiating StompAMQ. It shouldn't be the default for the code to guess which schema is being used. In case the validation happens in `make_notification` as in the previous point, users could specify which schema to use for each document, allowing for the possibility of one StompAMQ instance sending documents with several different schemas (since we already allow different doc types). We could set a default schema when instantiating StompAMQ and use that if none is provided. If none is provided, we could fall back to guessing.

- I think it should fall to the users to handle validation failures. I would suggest `make_notification` issues a logging.warning (or error?) when the validation fails and returns a bool with the result. Users can then keep track of documents that fail without dropping other documents in the same batch that pass validation.

Let me know what you think,
Benjamin

vkuznet:
Benjamin,
I agree with all of your suggestions, they seems reasonable to me. Would you
agree that if we move validate into `make_notification` then we should also
through the exception? I don't mind to add warning/error message, we just need
to decide which behavior we'll present to clients.

I'm still traveling and can't sit down and code, so feel free to make a PR,
otherwise I'll take care of this later this week.

Best,
Valentin.

stiegerb:
Ok, I'll prepare a PR, and we can iterate there.

I think the default behavior should be to return the notification and issue a warning or error in the logs. We can additionally offer to return also the result of the validation, so users could drop or collect those documents if they wanted.

In my opinion, the philosophy should be to check and notify users if their documents are failing validation, but to still send them, so to avoid data loss as much as possible.

stiegerb:
Ok, I opened a PR with some changes that I found to work for the job monitoring. Please have a look and let me know what you think.

One thing I was considering: maybe we should perform validation only on **known** keys and ignore unknown keys. Especially in the case of job monitoring, there are a lot of keys we might not care about, and we don't have much control over what is added in the future. Maintenance of this could be a serious headache.

stiegerb:
See also https://github.com/dmwm/cms-htcondor-es/pull/94

vkuznet:
[...]

PR #10, opened 2019-02-20
Handling of unknown fields

stiegerb:
As discussed in #8 
- Don't invalidate docs if they contain unknown fields, just issue a warning
- Increase the log level of failed validations to WARNING

### DIFF
--- a/src/python/CMSMonitoring/Validator.py
+++ b/src/python/CMSMonitoring/Validator.py
@@ -145,28 +145,28 @@ def _validate_schema(schema, doc):
 
     for key, val in doc.items():
         if key not in schema:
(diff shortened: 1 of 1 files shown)


vkuznet:
Benjamin, the patch is ok and I'll merge it, but I think the amount of logging info would be huge to handle. Remember that we use auto-validation of docs using default schemas, see
https://github.com/dmwm/CMSMonitoring/blob/46153c980155820013f6ff1ac3f7debfc71e5834/src/python/CMSMonitoring/StompAMQ.py#L50
Now, imaging that for particular doc we'll throw logging warnings from a schema which is not designed for this doc. All keys will be validated and we'll have N logging printouts for them. How can we distinguish which logging are true (because key is not present in schema) and which are fake (because we used wrong schema)?

PR #11, opened 2019-02-21
Change 'auto' behavior of validation, validation returns offending/unknown keys

stiegerb:
- When not choosing a validation schema, the validation now compares all docs to the first one by default (rather than trying out all possible schemas for every document). This will work ok for documents with a very fixed structure, i.e. where the first one is representative, but obviously goes wrong if documents are inhomogeneous or if the first one is a bad example.

- Users can now choose which log level the validation should use to issue feedback. In the same spirit, I removed the `verbose` toggle and replaced it with the log level.

- The validation now always runs on the entire document and collects a list of offending and unknown keys and then returns these. Added options to `make_notification` to drop these from the notification (either or both) and to return them to clients.

### DIFF
--- a/src/python/CMSMonitoring/StompAMQ.py
+++ b/src/python/CMSMonitoring/StompAMQ.py
@@ -207,8 +207,7 @@ class StompAMQ(object):
         return
 
     def make_notification(self, payload, docType, docId=None, producer=None, ts=None, metadata=None,
-                          dataSubfield="data", schema=None, returnOffendingKeys=False,
-                          dropOffendingKeys=False, dropUnknownKeys=False):
+                          dataSubfield="data", schema=None, dropOffendingKeys=False, dropUnknownKeys=False):
         """
         Produce a notification from a single payload, adding the necessary
         headers and metadata. Generic metadata is generated to include a
--- a/src/python/CMSMonitoring/StompAMQ.py
+++ b/src/python/CMSMonitoring/StompAMQ.py
@@ -50,9 +50,11 @@ def validate(doc, schema='auto', verbose=False):
         return validate_schema(_schemas.schemas()[schema], doc, verbose)
 
     else:
-        for sname, sch in _schemas.schemas().items():
-            if validate_schema(sch, doc, verbose):
-                return True
+        logging.warning('No validation schema provided, comparing all docs to the first one')
+        with open('auto_schema.json', 'w') as schfile:
+            json.dump(doc, schfile, indent=2, sort_keys=True)
+        _local_schemas['auto'] = doc
+        return True
 
     return False
 
--- a/src/python/CMSMonitoring/StompAMQ.py
+++ b/src/python/CMSMonitoring/StompAMQ.py
@@ -21,7 +21,7 @@ from CMSMonitoring.Validator import validate_schema, Schemas
 _schemas = Schemas(update=3600, jsonschemas=False)
 _local_schemas = None
 
-def validate(doc, schema='auto', verbose=False):
+def validate(doc, schema='auto', loglevel=logging.WARNING):
     """
     Helper function to validate given document against a schema
 
@@ -44,10 +44,10 @@ def validate(doc, schema='auto', verbose=False):
                 logging.warning('Local schema {} is not json compliant'.format(schema))
 
     if schema in _local_schemas:
-        return validate_schema(_local_schemas[schema], doc, verbose)
+        return validate_schema(_local_schemas[schema], doc, loglevel)
 
     elif schema in _schemas.schemas():
-        return validate_schema(_schemas.schemas()[schema], doc, verbose)
+        return validate_schema(_schemas.schemas()[schema], doc, loglevel)
 
     else:
         logging.warning('No validation schema provided, comparing all docs to the first one')
@@ -109,17 +109,19 @@ class StompAMQ(object):
         E.g.: [('agileinf-mb.cern.ch', 61213)]
     :param cert: path to certificate file
     :param key: path to key file
-    :param schema: schema to use for validation. If 'auto', will attempt to use central
-        CMSMonitoring schemas. If 'None', skip any validation. Look for schema files
-        locally, in 'schemas/' folder in CMSMonitoring package or in folder defined in
-        'CMSMONITORING_SCHEMAS' environmental variable.
+    :param validation_schema: schema to use for validation (filename of a valid json file).
+        If 'auto', will compare all docs to the first one provided. If 'None', skip any
+        validation. Look for schema files locally, in 'schemas/' folder in CMSMonitoring
+        package or in folder defined in 'CMSMONITORING_SCHEMAS' environmental variable.
+    :param validation_loglevel: logging level to use for validation feedback
     """
 
     # Version number to be added in header
     _version = '0.3'
 
     def __init__(self, username, password, producer, topic,
-                 host_and_ports=None, logger=None, cert=None, key=None, schema='auto'):
+                 host_and_ports=None, logger=None, cert=None, key=None,
+                 validation_schema='auto', validation_loglevel=logging.WARNING):
         self._username = username
         self._password = password
         self._producer = producer
@@ -132,7 +134,8 @@ class StompAMQ(object):
         # silence the INFO log records from the stomp library, until this issue gets fixed:
         # https://github.com/jasonrbriggs/stomp.py/issues/226
         logging.getLogger("stomp.py").setLevel(logging.WARNING)
-        self.schema = schema
+        self.validation_schema = validation_schema
+        self.validation_loglevel = validation_loglevel
 
     def send(self, data):
         """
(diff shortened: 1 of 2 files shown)


stiegerb:
Adding also the change in return value here, as the two branches would conflict anyway.

src/python/CMSMonitoring/StompAMQ.py
```diff
@@ -44,17 +47,19 @@ def validate(doc, schema='auto', verbose=False):
...
 
     elif schema in _schemas.schemas():
-        return validate_schema(_schemas.schemas()[schema], doc, verbose)
+        return validate_schema(_schemas.schemas()[schema], doc, loglevel)
 
     else:
-        for sname, sch in _schemas.schemas().items():
-            if validate_schema(sch, doc, verbose):
-                return True
+        logging.warning('No validation schema provided, comparing all docs to the first one')
```
vkuznet: This part will not work in general. Here is an example using WMAgent:
- client didn't provide a schema
- it forces to create a file `auto_schema.json`
- then client validate doc1 which confirm with this schema
- but then client use another doc which does not belong to the schema, e.g. WMA may have different type of docs, according to Alan they may have around 30 different schemas. At this point the validation will fail since it relies on doc1 schema which may not be applicable to other docs.

I also dislike creation of `auto_schema.json` files at random place (this is what code implies) since such documents will be created at run time environment there is no procedure to clean-up them.

There is also no connection how to put this schema back into repository and therefore we will not have track of what has been validated. Having schemas in one repository which we can tag provides us with a way to trace down which schemas (versions) was used in particular period of time.

My suggestion to get rid of auto complete and force users to specify the schema name. All schemas will be part of repository and API should either provide existing file name (which we can check with `os.path.exists`) or file name which should match existing schemas from repository.

src/python/CMSMonitoring/StompAMQ.py
```diff
@@ -44,17 +47,19 @@ def validate(doc, schema='auto', verbose=False):
...
+        with open('auto_schema.json', 'w') as schfile:
+            json.dump(doc, schfile, indent=2, sort_keys=True)
+        _local_schemas['auto'] = doc
+        return [], []
```
vkuznet: Another issue, you're mixing return data-types, here it is list, list, while other returns None, None. It is better to have consistent return data-type.

src/python/CMSMonitoring/StompAMQ.py
```diff
@@ -270,6 +292,6 @@ def make_notification(self, payload, docType, docId=None, producer=None, ts=None
         notification.update(headers)
         notification['body'] = body
 
-        if returnValidationResult:
-            return notification, validated
+        if returnOffendingKeys:
+            return notification, offending_keys, unknown_keys
```
vkuznet: Here you return two lists in addition to notification, while on line below it is only notification. Usually different return metrics may cause more trouble upstream since clients need to know how may return parameters to expect and their code may break depending on run-time conditions. For example, let's say client call is
```
    returnOffendingKeys = config.get() # client get its parameter at run time
    notification = make_notification(..., returnOffendingKeys)
```
this code will not behave as expecting if `returnOffendingKeys` parameter change at run-time. Since when it is not set code will return one notification parameter, while when it is set it will get three parameters.

It is much safer to have consistent set of return parameters which consistent data-types.

src/python/CMSMonitoring/StompAMQ.py
stiegerb: As I wrote above, this is meant as a somewhat more sensible default behavior, but obviously users should be providing a schema and not rely on this. I'm happy to remove it altogether.

On the other hand, it doesn't really make a difference. Someone using this will quickly run into validation failures and can then create a schema and use that.

If you insist, I'm happy to remove it altogether.

src/python/CMSMonitoring/StompAMQ.py
stiegerb: You have a better suggestion? The point is to distinguish between the case where the validation was successful and the one where none was run. Note that the `None, None` only appears if people don't run validation. In that case they'll hopefully not use the validation result in the first place.

src/python/CMSMonitoring/StompAMQ.py
stiegerb: Agreed, this isn't very pretty. We introduced the `returnValidationResult` to be backwards compatible, but I'm happy to remove it and just return the lists in any case.

vkuznet:
Benjamin, I think solution is trivial:
- get rid of `returnValidationResult` option and always return results, if validation successful user will get back empty lists, if not it will get the ids
- remove auto mode and enforce users to specify the schema
- the schema should ONLY be taken from internal area (of CMSMonitoring/schema or CMSMonitoring/jsonschema), clients should not allowed to create schema with the docs. This will solve your concern about making schemas. If schema is not found we return that validation failed. That's why it is better to introduce validation codes, 0 means success, 1 means keys are not validated in internal schema, 2 means that schema is not found, etc. We can introduce codes in a code. The client may check the code to understand what's wrong.
- we will accept clients schemas and add them to repository manually. This way we have full control on schemas and can version them.

Can you do this in this PR? or do you want to break this down into different pieces?

stiegerb:
- As I wrote, I'm happy to get rid of the `returnValidationResult` option. I'll update the PR later. But note that `make_notification` cannot return lists of ids as it applies to **single documents**. The lists contain the offending and unknown keys from the validation of the one document passed to the function.

- Since you insist, I'll remove the 'auto' mode.

- I'm afraid I don't understand what you mean in your third point. You want to remove the possibility of using local schema files? Which is my concern about making schemas? Where do you want to put these validation codes? Maybe it's more efficient if we have a quick chat?

stiegerb:
Ok, quick summary of our chat:
- We'll remove the 'auto' option and return `None, None` in case no schema is provided.
- `make_notification` checks if the return value of `validate` is `None` and issues an error message.
- The code looks for schemas first locally, then in the CMSMonitoring/schemas locations (already the case)
- The user-provided schema name is checked against filenames with and without the .json extension
- The `validation_schema` parameter of StompAMQ becomes required, but users can set it to `None` if they want to skip validation.
- In the future, we will need to update the code such that the functionality of returning lists of offending and unknown keys works also with the jsonschema branch

@vkuznet let me know if I missed something. I'll update the PR in the meantime.

vkuznet:
+1

stiegerb:
Ok, I update the PR with these changes. Have a look. If you're happy, I can squash and force-push, or you can squash and merge.

stiegerb:
(Only one difference to what we discussed. I think it's cleaner to have `validate` issue the error message if it doesn't find the schema, rather than having `make_notification` check the return value first.)

src/python/CMSMonitoring/StompAMQ.py
```diff
@@ -41,20 +42,18 @@ def validate(doc, schema='auto', verbose=False):
...
-    elif schema in _schemas.schemas():
-        return validate_schema(_schemas.schemas()[schema], doc, verbose)
+        return validate_schema(_local_schemas[schema], doc, loglevel)
 
     else:
-        for sname, sch in _schemas.schemas().items():
-            if validate_schema(sch, doc, verbose):
-                return True
+        for sch in _schemas.schemas():
+            if schema in [sch, os.path.basename(sch).rsplit('.')[0]]:
```
vkuznet: I would replace it with
```
if sch.find(schema) != -1
```
which basically string operation on sch name to find a sub-string schema in it.

src/python/CMSMonitoring/StompAMQ.py
```diff
@@ -202,7 +207,8 @@ def _send_single(self, conn, notification):
         return
 
     def make_notification(self, payload, docType, docId=None, producer=None, ts=None, metadata=None,
-                          dataSubfield="data", schema=None, returnValidationResult=True):
+                          dataSubfield="data", schema=None, returnOffendingKeys=False,
```
vkuznet: I think we agreed to change default behavior and use schema as mandatory parameter. If users want to skip it they should explicitly provide schema=None. But since schema will be mandatory parameter it should come before optional parameter, i.e. be moved before docId.

src/python/CMSMonitoring/StompAMQ.py
stiegerb: That would be equivalent to `if schema in sch`, no? I went with the explicit approach because otherwise you get tricky bugs when people name their schemas "JobMonitoring.json" and "JobMonitoringExtra.json" and the "JobMonitoring" string will hit both.

src/python/CMSMonitoring/StompAMQ.py
stiegerb: That's a different place in the code. The required parameter is in the constructor of StompAMQ. Here in `make_notification` the default behavior (i.e. giving `None`) is to fall back to the schema provided when setting up StompAMQ. I think it's ok as it is.

But highlighting this made me notice that I forgot to remove the `returnOffendingKeys` parameter from the function declaration. I'll add another commit to fix it.

vkuznet:
ok, merging.

PR #34, opened 2019-08-01
Use the external datasources list

cronosnull:
Instead of a dictionary, we want to use the list generated with datasourcesList.py (from the grafana data sources)

This let us to several changes:

- An option to list the data sources
- Instead of the database name, we require the name of the data source in grafana. 
- we can use the type field in the list to determine the type of the query (influxdb/elasticsearch). 
- the python script and the monit bin expect the file to be at the static/datasources.json path (configurable via env var)

(tested using python3)

### DIFF
--- a/bin/monit
+++ b/bin/monit
@@ -1,6 +1,8 @@
 #!/bin/sh
 
 # find out where CMSMonitoring is installed on a system
-root=`python -c "import CMSMonitoring; print '/'.join(CMSMonitoring.__file__.split('/')[:-1])"`
+root=$(python -c "import CMSMonitoring; print('/'.join(CMSMonitoring.__file__.split('/')[:-1]))")
+MONIT_HOME="$(cd "$(dirname "$0")" && pwd)"
+export MONIT_DB_DICT_FILE="$MONIT_HOME/../static/datasources.json"
 # run actual script
 python $root/monit.py ${1+"$@"}
--- a/bin/monit
+++ b/bin/monit
@@ -3,6 +3,6 @@
 # find out where CMSMonitoring is installed on a system
 root=$(python -c "import CMSMonitoring; print('/'.join(CMSMonitoring.__file__.split('/')[:-1]))")
 MONIT_HOME="$(cd "$(dirname "$0")" && pwd)"
-export MONIT_DB_DICT_FILE="$MONIT_HOME/../static/datasources.json"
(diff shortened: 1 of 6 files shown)


bin/monit
```diff
@@ -1,6 +1,8 @@
 #!/bin/sh
 
 # find out where CMSMonitoring is installed on a system
-root=`python -c "import CMSMonitoring; print '/'.join(CMSMonitoring.__file__.split('/')[:-1])"`
+root=$(python -c "import CMSMonitoring; print('/'.join(CMSMonitoring.__file__.split('/')[:-1]))")
+MONIT_HOME="$(cd "$(dirname "$0")" && pwd)"
+export MONIT_DB_DICT_FILE="$MONIT_HOME/../static/datasources.json"
```
vkuznet: This is fragile, instead do something like this:
```
MONIT_DB_DICT_FILE=${MONIT_DB_DICT_FILE:-$MONIT_HOME/../static/datasources.json}
```
This will allow to specify datasources.json via environment variable and have fallback to MONIT home area (or whatever).

src/python/CMSMonitoring/datasourcesList.py
```diff
@@ -8,54 +8,70 @@
...
-        self.parser = argparse.ArgumentParser(prog='PROG', usage=desc)
-        self.parser.add_argument("--token", action="store",
-            dest="token", default=None, help="Admin token")
-        self.parser.add_argument("--url", action="store",
-            dest="url", default="https://monit-grafana.cern.ch", help="MONIT URL")
-        self.parser.add_argument("--output", action="store",
-            dest="output", default=sys.stdout, help="output file")
-
-def get_datasources(token, base='https://monit-grafana.cern.ch'):
+        self.parser = argparse.ArgumentParser(prog="PROG", usage=desc)
```
vkuznet: you need to define desc here. I suggest to put into description a note about all environment variables this script relies on along with option descriptions.

bin/monit
cronosnull: Agree :+1:

src/python/CMSMonitoring/datasourcesList.py
cronosnull: It is defined, just doesn't change from the current version (I run black to format the code and it touch some lines for style fixes), but I agree that it should be more explanative and include the definition of the env vars. I added that in the 
845fa15  commit.

PR #36, opened 2019-08-07
Popularityplot - DRAFT

cronosnull:
The script to create the scrutiny plots from hdfs data. 

Currently it use the dbs_condor dataset from hdfs, but still uses the local phedex and events.tgz files in vocms092. The objective is to use hdfs as the only source (so we can put this script as an acron in lxplus, for example, instead of a cronjob in vocms092 as it currently is)

### DIFF
--- a/src/python/CMSMonitoring/ScrutinyPlot.py
+++ b/src/python/CMSMonitoring/ScrutinyPlot.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 
 # ## Params
-# start dates of each interval and end date. 
+# start dates of each interval and end date.
 
 # In[1]:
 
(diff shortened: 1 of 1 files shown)


vkuznet:
Christian, could you please re-factor code such that you'll put all global code logic into proper functions, i.e. avoid writing code in global namespace since if we'll need to import the module all of these globals will be exposed. I usually follow up this logic in every code:
```
#!/usr/bin/env python
#-*- coding: utf-8 -*-
#pylint: disable=
"""
File       : ttt.py
Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
Description:
"""

# system modules
import os
import sys
import argparse

class OptionParser():
    def __init__(self):
        "User based option parser"
        self.parser = argparse.ArgumentParser(prog='PROG')
        self.parser.add_argument("--fin", action="store",
            dest="fin", default="", help="Input file")
        self.parser.add_argument("--fout", action="store",
            dest="fout", default="", help="Output file")
        self.parser.add_argument("--verbose", action="store_true",
            dest="verbose", default=False, help="verbose output")

def main():
    "Main function"
    optmgr  = OptionParser()
    opts = optmgr.parser.parse_args()

if __name__ == '__main__':
    main()
```

cronosnull:
It nows uses only data from HDFS, and can run as an "acron" job in lxplus

cronosnull:
@vkuznet  I think it is ready for review.

PR #37, opened 2019-08-13
Grafana duplicator

cronosnull:
It can use an *uid* or a *query* expression to select the dashboards to copy, creates the documents to create new dashboards, changing the title and replacing the datasources if needed (in all panels and variables). 

It's a draft because it needs some additional testing and documentation.

### DIFF
--- a/doc/scripts/GrafanaDuplicator.md
+++ b/doc/scripts/GrafanaDuplicator.md
@@ -1,0 +1,36 @@
+# Grafana duplicator
+
+[This app](../../src/python/CMSMonitoring/grafana_duplicator.py) allows to create a copy of one (using the uid parameter) or a set of dashboards (using the query parameter) allowing to change the datasource and the id and the title. 
+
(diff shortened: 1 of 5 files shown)


cronosnull:
I have separated the Grafana_manager class from the duplicator script and added a text propagation script that uses it too. 
It can be useful to update the tabs on the grafana homepage (now that we have more than 20 tabs). The important thing when adding a new tab is tagging the new dashboard with "home", adding the new link in the HomeTabs dashboard and then run the script to update the menu in all the other "home" dashboards. 

https://github.com/cronosnull/CMSMonitoring/blob/grafana_duplicator/doc/scripts/GrafanaTextPropagation.md

vkuznet:
Christian, does this PR is ready or do I need to wait for more changes?

cronosnull:
I just added some missing documentation, now I think it is ready to merge. 

Cheers,

PR #51, opened 2020-05-18
Daemon script + config added

indrarahul2013:
Script for daemon service added which will be querying CERN SSB Status Board endpoint and saves the output and then our alerting tool will inject them into AlertManager.


vkuznet:
Rahul, a few things:
- why do you need to execute `go run` every time you need to call the script? Why not rely on the Unix environment which can be setup up-front to point to the path to use compiled executables `monit` and `alerting`?
- we don't need separate `config` file, instead, the bash script should take input arguments to start-up
- you should not use eval to execute shell commands, please read documentation why and how eval is used
- you should not place shell script into src/go area which is designed for go source code, the scripts should be placed into CMSMonitoring/scripts area.

Basically what you need is this:
```
#!/bin/bash
# check that user give 3 arguments
if [ "$#" -eq 3 ]; then
    echo "ssb_alerts.sh provides basic daemon to fetch SSB/GGUS info and inject it into MONIT"
    echo "Usage: ssb_alerts.sh <query> <token-file> <interval (in sec)>"
    exit 1
fi

query=$1
token=$2
interval=$3

# Alerting Tool arguments
input_file=data.json
alertmanager_url='http://localhost:9093'
verbose=2

while true;  do
   monit -query="$query" -dbname=monit_production_ssb_otgs -token=$token -dbid=9474 > $ssb_file
   alerting -input $ssb_file -url $alertmanager_url -verbose $verbose
   sleep $interval
done 
```
Please read/learn how to write proper shell scripts to handle input arguments. For instance, I already provided example above how to check input arguments and provide basic help, etc.

vkuznet:
please expand the provided example to properly configure it, e.g. add additional argument to specify alert-manager host, etc. Then, test it and confirm that it works. For tests, you should compile your monit and alerting executable, setup your UNIX PATH to include their location, etc.

scripts/ssb_alert.sh
```diff
@@ -0,0 +1,23 @@
...
+#Check if user is passing all arguments.
+if [ "$#" != 5  ]; then
+    echo "ssb_alerts.sh provides basic daemon to fetch CERN SSB info and inject it into MONIT (AlertManager) "
+    echo "Usage: ssb_alerts.sh <ssb_query> <token-file> <alertmanager-url> <interval (in sec)> <verbosity level>"
+    exit 1
+fi
+
+query=$1
+token=$2
+alertmanager_url=$3
```
vkuznet, line 12: The url, interval and verbose should have proper defaults, e.g.
```
url=${3:https//cms-monitoring.cern.ch}
interval=${4:-1}
verbose=${5:-0}
```
such that we don't need to specify them all the time. But it implies that you should properly adjust how to check number of input arguments, adjust usage string, etc.

Please adjust accordingly and test that it will work with and without defauls.

indrarahul2013:
Valentin,
Thanks for the detailed explanation. I have addressed all issues. I added few more arguments in the script such as alertmanager-url, verbosity level for alerting tool.

scripts/ssb_alert.sh
indrarahul2013: Solved this issue.

indrarahul2013:
Linux Service added.

scripts/ssb_alert_manage
```diff
@@ -0,0 +1,89 @@
...
+
+ME=$(basename $(dirname $0))
+TOP=$(cd $(dirname $0)/../../.. && pwd)
+ROOT=$(cd $(dirname $0)/../.. && pwd)
+LOGDIR=$TOP/logs/$ME
+COLOR_OK="\\033[0;32m"
+COLOR_WARN="\\033[0;31m"
+COLOR_NORMAL="\\033[0;39m"
+
+#. $ROOT/apps/$ME/etc/profile.d/init.sh
```
vkuznet, line 24: Remove if not used

scripts/ssb_alert_manage
```diff
@@ -0,0 +1,89 @@
...
+
+### ***
+
+# Actions Logic
+start()
+{
+    echo "Starting Alerting daemon."
+    nohup ssb_alert.sh $query $token </dev/null 2>&1 | rotatelogs $LOGDIR/ssb_alerting-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
```
vkuznet, line 32: what will happen if system does not have rotatelogs executable? Where $query and $token are defined? How to pass additional parameters to ssb_alert.sh?

You may pass query/token and other parameters via environment variables to manage script, but please make that they will have valid defaults. Then pass all parameters to ssb_alert.sh

The bottom line, please check that everything works and adjust manage script to work with and without rotatelogs. Check that $LOGDIR is actually present on a system and manage script has permission to write to it

indrarahul2013: Valentin,

I have implemented checks for rotatelogs and $LOGDIR. Regarding arguements for ssh_alert.sh, I assumed enivornment variables only in the last commit as well, but all the arguements were not present (optional args). In latest commit this has been implemented - 
```
## ssb_alert.sh arguments from Environment Variable
QUERY=$query                                  ## monit influxDB query
TOKEN=$token                                  ## token file
URL=${url:-"https://cms-monitoring.cern.ch"}   ## alertmanager_url
INTERVAL=${interval:-1}                       ## time interval at which fetching and injecting data repeats
VERBOSE=${verbose:-0} 
```
Where query, token are required environment variables and url,interval,verbose are optional. Here they will have default values if not provided explicitly.

vkuznet:
Rahul, please NEVER upload binaries to github, so far I see that you committed bin/monit, bin/alerting, etc. executable. Please delete them from PR and your repository.

If you need to keep them around please use .gitignore to specify what git should ignore (google it).

indrarahul2013:
Valentin,
I have deleted the binaries uploaded before, actually the monit binary was already in the dmwm/CMSMonitoring/bin repository. The new commit will fix this issue as well. 

Thanks

vkuznet:
Rahul, the CMSMonitoring/bin/monit is not a binary, it is shell script to run python code. Therefore, please put it back in this PR. I'm sorry I overlooked this, what I was talking that you had real executables (e.g. alerting) committed and we should not do that.

Please let me know again when your PR will be ready to review.

indrarahul2013:
Valentin,
I have added the monit binary back. Please review.

scripts/ssb_alert_manage
```diff
@@ -0,0 +1,109 @@
...
+INTERVAL=${interval:-1}                       ## time interval at which fetching and injecting data repeats
+VERBOSE=${verbose:-0}                         ##verbosity level
+
+# Actions Logic
+start()
+{
+    echo "Starting Alerting daemon."
+    if [ ! -x "$(command -v rotatelogs)" ]; then
+      echo "rotatelogs not found. Starting without it..."
+      nohup ssb_alert.sh "$QUERY" $TOKEN $URL $INTERVAL $VERBOSE </dev/null 2>&1 &
```
vkuznet, line 49: we still need to keep the log if we don't have rotatelogs, please redirect to logdir in a similar way as you you have below with rotate log.

bin/monit
```diff
@@ -5,4 +5,4 @@ root=$(python -c "import CMSMonitoring; print('/'.join(CMSMonitoring.__file__.sp
 MONIT_HOME="$(cd "$(dirname "$0")" && pwd)"
 export MONIT_DB_DICT_FILE="${MONIT_DB_DICT_FILE:-$MONIT_HOME/../static/datasources.json}"
 # run actual script
-python "$root/monit.py" ${1+"$@"}
+python "$root/monit.py" ${1+"$@"}
```
vkuznet, line 8: there is no need to modify the file, leave it intact.

indrarahul2013: Valentin,
The monit binary looks same. I am not sure why any change appeared.

vkuznet: it is your editor which may add newline, etc. So any file you can revert to previous form by using `git checkout -- <file_name>`. But this is minor detail, I can merge the PR once you'll fix the log

indrarahul2013: Valentin,
I have updated both files. Please review. Thanks for the git help, i will keep in mind next time.

scripts/ssb_alert_manage
indrarahul2013: Fixed this issue. Please review.

.gitignore
```diff
@@ -84,6 +84,9 @@ celerybeat-schedule
 # SageMath parsed files
 *.sage.py
 
+# bin directory
+bin/
```
vkuznet, line 88: This will ignore bin/monit script, I didn't ask for this feature, we can still keep in bin area some scripts, what we don't want is to keep executables over there. Please remove it

scripts/ssb_alert_manage
```diff
@@ -0,0 +1,109 @@
...
+LOGDIR=$TOP/logs/$ME
+COLOR_OK="\\033[0;32m"
+COLOR_WARN="\\033[0;31m"
+COLOR_NORMAL="\\033[0;39m"
+
+### ***
+
+if [ ! -d $LOGDIR ]; then 
+  echo "Log Directory '$LOGDIR' does not exists. Creating..."
+  mkdir -p $LOGDIR
```
vkuznet, line 28: this will fail if you don't have permission to create LOGDIR, please check first that you have permission over there.

scripts/ssb_alert_manage
```diff
@@ -0,0 +1,109 @@
...
+INTERVAL=${interval:-1}                       ## time interval at which fetching and injecting data repeats
+VERBOSE=${verbose:-0}                         ##verbosity level
+
+# Actions Logic
+start()
+{
+    echo "Starting Alerting daemon."
+    if [ ! -x "$(command -v rotatelogs)" ]; then
+      echo "rotatelogs not found. Starting without it..."
+      nohup ssb_alert.sh "$QUERY" $TOKEN $URL $INTERVAL $VERBOSE </dev/null 2>&1 | tee $LOGDIR/ssb_alerting-`date +%F`-`hostname -s`.log >/dev/null 2>&1 &
```
vkuznet, line 49: why do you need `tee` to keep output on stdout? Please redirect directly to log file, that's it. If someone needs to see a progress than this person can use `tail -f <file.log>` for that. Basically you need to have the following:
```
nohup ssb_alert.sh "$QUERY" $TOKEN $URL $INTERVAL $VERBOSE 2>&1 1>& $LOGDIR/ssb_alerting-`date +%F`-`hostname -s`.log &
```

indrarahul2013: My understanding was to redirect the stdout data to log file using `tee`. Thanks for the simpler approach. I have updated the line.

scripts/ssb_alert_manage
indrarahul2013: Right, I have now implemented the check for `$TOP` directory. I have also put check for `mkdir` command if it returns success or not.

.gitignore
indrarahul2013: Thanks, I will keep this in mind.

indrarahul2013:
> OK, now I see that your PR contains empty file `scripts/now`, what is that? Please remove it.

Removed. I apologize for such mistake, will take care not to create unwanted files.

PR #52, opened 2020-05-21
Added ggus_alerting files, updated ssb_alerting files

indrarahul2013:
Implementation of alerting tool for the GGUS Ticketing System which will ingest JSON data and inject them into AlertManager. Ending Time for each alerts has been implemented on the logic of checking if a particular ticket has been resolved or not, and updating the same in AlertManager.

### DIFF
--- a/src/go/MONIT/ggus_alerting.go
+++ b/src/go/MONIT/ggus_alerting.go
@@ -1,0 +1,357 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"os"
+	"strconv"
+	"time"
+)
+
+// File       : ggus_alerting.go
+// Author     : Rahul Indra <indrarahul2013 AT gmail dot com>
+// Created    : Thu, 21 May 2020 21:07:32 GMT
+// Description: GGUS Alerting Module for CERN MONIT infrastructure
+
+//URL for AlertManager
+var alertManagerURL string
+
+//severity of alerts
+var severity string
+
+//GGUS tag
+var tag string
+
+//verbose defines verbosity level
+var verbose int
+
+//Map for storing Existing Tickets
+var exstTkt map[string]int
+
+//GGUS Data Struct
+type ggus []struct {
+	TicketID        int    `json:"TicketID"`
+	Type            string `json:"Type"`
+	VO              string `json:"VO"`
+	Site            string `json:"Site"`
+	Priority        string `json:"Priority"`
+	ResponsibleUnit string `json:"ResponsibleUnit"`
+	Status          string `json:"Status"`
+	LastUpdate      int    `json:"LastUpdate"`
+	Subject         string `json:"Subject"`
+	Scope           string `json:"Scope"`
+}
+
+//AlertManager API acceptable JSON Data for GGUS Data
+type amJSON struct {
+	Labels struct {
+		Alertname string `json:"alertname"`
+		Severity  string `json:"severity"`
+		Tag       string `json:"tag"`
+	} `json:"labels"`
+	Annotations struct {
+		TicketID        string `json:"TicketID"`
+		Type            string `json:"Type"`
+		VO              string `json:"VO"`
+		Site            string `json:"Site"`
+		Priority        string `json:"Priority"`
+		ResponsibleUnit string `json:"ResponsibleUnit"`
+		Status          string `json:"Status"`
+		Subject         string `json:"Subject"`
+		Scope           string `json:"Scope"`
+		URL             string `json:"URL"`
+	} `json:"annotations"`
+	StartsAt time.Time `json:"startsAt"`
+	EndsAt   time.Time `json:"endsAt"`
+}
+
+//AlertManager GET API acceptable JSON Data struct for GGUS data
+type getAMJSON struct {
+	Status string `json:"status"`
+	Data   []struct {
+		Labels struct {
+			Alertname string `json:"alertname"`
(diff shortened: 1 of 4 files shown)


src/go/MONIT/ggus_alerting.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		panic(err)
+	}
+	defer resp.Body.Close()
+
+	byteValue, err := ioutil.ReadAll(resp.Body)
```
vkuznet, line 215: Will it work for alerts which will have different attributes then ggus ones?

src/go/MONIT/ggus_alerting.go
```diff
@@ -0,0 +1,357 @@
...
+		Subject         string `json:"Subject"`
+		Scope           string `json:"Scope"`
+		URL             string `json:"URL"`
+	} `json:"annotations"`
+	StartsAt time.Time `json:"startsAt"`
+	EndsAt   time.Time `json:"endsAt"`
+}
+
+//AlertManager GET API acceptable JSON Data struct for GGUS data
+type getAMJSON struct {
```
vkuznet, line 75: What is the reason to name a struct with *get* prefix? The get prefix indicates that you get the data, so it is more appropriate for a function name. Please rename it appropriately, e.g. ggusData.

src/go/MONIT/ggus_alerting.go
```diff
@@ -0,0 +1,357 @@
...
+	ResponsibleUnit string `json:"ResponsibleUnit"`
+	Status          string `json:"Status"`
+	LastUpdate      int    `json:"LastUpdate"`
+	Subject         string `json:"Subject"`
+	Scope           string `json:"Scope"`
+}
+
+//AlertManager API acceptable JSON Data for GGUS Data
+type amJSON struct {
+	Labels struct {
```
vkuznet, line 53: This part repeats in getAMJSON struc, therefore take it out and re-use ,e.g.
```
type Labels struct {
  ...
}
type amJSON struct {
....
   Labels Labels
...
}
```

src/go/MONIT/ggus_alerting.go
```diff
@@ -0,0 +1,357 @@
...
+		Alertname string `json:"alertname"`
+		Severity  string `json:"severity"`
+		Tag       string `json:"tag"`
+	} `json:"labels"`
+	Annotations struct {
```
vkuznet, line 58: Similar here, the Annotations is identical in both structs, in amJSON and in getAMJSON. You should define it outside of them and then use it.

src/go/MONIT/ggus_alerting.go
```diff
@@ -0,0 +1,357 @@
...
+
+	var temp amJSON
+	var finalData []amJSON
+
+	exstTkt = make(map[string]int)
+
+	for _, each := range data {
+
+		// Ignoring all tickets whose VO is not "cms"
+		if each.VO != "cms" {
```
vkuznet, line 149: Make it configurable, such that if we will need to change it later we'll have this flexibility, i.e. we should be able to pass filter to the program, but it should have default value "cms".

src/go/MONIT/ggus_alerting.go
indrarahul2013: Yes ! I have injected both alerts (SSB & GGUS Data) and it works independently. Please let me know if I still have to filter the GGUS data.

src/go/MONIT/ggus_alerting.go
indrarahul2013: Fixed this issue.

src/go/MONIT/ggus_alerting.go
indrarahul2013: Fixed the getAMJSON struct. Named to ggusData, removed unused variables.

src/go/MONIT/ggus_alerting.go
indrarahul2013: Fixed this.

src/go/MONIT/ggus_alerting.go
indrarahul2013: Renamed it to ggusData.

src/go/MONIT/ggus_alerting.go
```diff
@@ -0,0 +1,332 @@
...
+		VO              string `json:"VO"`
+		URL             string `json:"URL"`
+	} `json:"annotations"`
+	StartsAt time.Time `json:"startsAt"`
+	EndsAt   time.Time `json:"endsAt"`
+}
+
+//AlertManager GET API acceptable JSON Data struct for GGUS data
+type ggusData struct {
+	Data []amJSON
```
vkuznet, line 79: If ggusData is really a list of amJSON then why do you need this struct at all? You can easily Unmarshal data using `[]amJSON`.

indrarahul2013: Valentin,
When I call /api/v1/alerts (GET) API, I get struct of data array from AlertManager. It is of form :- 
```
{
   "data" : [ {},
              {},
             ],
. . . 
}
```

So, I would be requiring struct to parse the information.

vkuznet: ok, then it is correct.

src/go/MONIT/ggus_alerting.go
indrarahul2013: I have added the filter for fetching only GGUS alerts from AlertManager. 
```
apiurl := alertManagerURL + "/api/v1/alerts?active=true&silenced=false&inhibited=false&unprocessed=false&filter=tag=\"GGUS\""
```

PR #58, opened 2020-05-25
Fixed ggus_alerting.go

indrarahul2013:


src/go/MONIT/ggus_alerting.go
```diff
@@ -273,9 +277,15 @@ func deleteAlerts() {
 		temp.Annotations.Scope = each.Annotations.Scope
 		temp.Annotations.URL = each.Annotations.URL
 
-		temp.StartsAt = each.StartsAt
+		emp.StartsAt = each.StartsAt
```
vkuznet, line 280: are you sure that `emp` is an appropriate name here?

indrarahul2013: No, it was a mistake.

vkuznet: next time try to compile your code before submitting the PR since the compiler will fetch this kind of typos.

indrarahul2013: Yes. I will follow this. 
Thanks

PR #64, opened 2020-06-15
alert CLI Tool - Flag Usage updated

indrarahul2013:
@vkuznet  alert CLI Tool - Flag Usage updated


src/go/MONIT/alert.go
```diff
@@ -576,31 +576,31 @@ func main() {
...
+		fmt.Println("\tGet all alerts:")
+		fmt.Println("\t    alert")
+		fmt.Println("\tGet all alerts in JSON format:")
+		fmt.Println("\t    alert -json")
+		fmt.Println("\tGet all alerts with filters (-json flag will output in JSON format if required):")
+		fmt.Println("\t    Available filters:")
+		fmt.Println("\t    service\tEx GGUS,SSB,dbs,etc.")
+		fmt.Println("\t    severity\tEx info,medium,high,urgent,etc.")
+		fmt.Println("\t    tag\t\tEx cmsweb,cms,monitoring,etc.")
+		fmt.Println("\n\t    Get all alerts of specific service/severity/tag. Ex GGUS/high/cms:")
```
vkuznet, line 588: I'm not sure if you need additional tag/spaces here. I would suggest that you align all "Get" messages on the same column.

indrarahul2013: Valentin, 
The tabbed "Get" messages are sub-messages. Should I still align them in same column?

vkuznet: All you gets, are just examples, do not complicate things where it is not necessary, it only leads to confusion from user side

indrarahul2013: Okay. I have made changes.

indrarahul2013:
@vkuznet, now you can merge.

PR #65, opened 2020-06-26
Intelligence Module added.

indrarahul2013:


indrarahul2013:
Things to be done -
- Flag Information to be updated.
- systemd file to be added.

src/go/MONIT/intelligence.go
```diff
@@ -0,0 +1,320 @@
...
+	for true {
+		run()
+		time.Sleep(configJSON.Interval * time.Second)
+	}
+}
+
+func main() {
+
+	var verbose int
+	flag.IntVar(&verbose, "verbose", 0, "Verbosity Level, can be overwritten in config")
```
vkuznet, line 308: make config option to pass config file which will replace reading it from environment

src/go/MONIT/intelligence.go
```diff
@@ -0,0 +1,320 @@
...
+	get(&amdata)
+
+	filtered := filterMaintenance(amdata)
+	silenceMaintenance(filtered)
+
+}
+
+func parseConfig(verbose int) {
+
+	configFile := os.Getenv("INTELLIGENCE_CONFIG_PATH") //INTELLIGENCE_CONFIG_PATH Environment Variable storing config filepath.
```
vkuznet, line 256: pass configFile as input argument to this function

src/go/MONIT/intelligence.go
```diff
@@ -0,0 +1,320 @@
...
+	configJSON.SeverityFilter = "maintenance"
+	configJSON.SearchingLabel = "instance"
+	configJSON.UniqueLabel = "alertname"
+	configJSON.Comment = "maintenance"
+	configJSON.CreatedBy = "admin"
+	configJSON.HttpTimeout = 3 //3 secs timeout for HTTP requests
+	configJSON.Interval = 10   // 10 sec interval for the service
+	configJSON.Verbose = verbose
+
+	if configFile != "" {
```
vkuznet, line 271: use os.stats to check if configFile is indeed exists, eg.
```
if stats, err := os.Stat(fname); if err == nil {
} else { // if file does not exists provide some loving info
   log.Printf("provided config file %s does not exists, error %v\n", configFile, err)
}
```

src/go/MONIT/intelligence.go
```diff
@@ -0,0 +1,320 @@
...
+	sData.StartsAt = mData.StartsAt //Start Time equal to maintenance alert	   //So that when maintenance alerts vanishes, ongoing alerts comes back alive.
+	sData.EndsAt = mData.EndsAt     //End Time equal to maintenance alert	//So that when maintenance alerts vanishes, ongoing alerts comes back alive.
+	sData.CreatedBy = configJSON.CreatedBy
+	sData.Comment = configJSON.Comment
+	var m matchers
+
+	m.Name = configJSON.UniqueLabel
+	for k, v := range data.Labels {
+		if k == configJSON.UniqueLabel {
+			m.Value = v.(string)
```
vkuznet, line 149: You need to use switch clause to properly cast data types, what if v will be nil and you're trying to cast it to string?

src/go/MONIT/intelligence.go
```diff
@@ -0,0 +1,320 @@
...
+		if err == nil {
+			log.Println("Request: ", string(dump))
+		}
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		panic(err)
+	}
+	defer resp.Body.Close()
```
vkuznet, line 177: Can you check response status and make function return an error instead of panic above. In the case of 200 HTTP status you'll return nil, otherwise you'll construct proper error which upstream code can handle.

src/go/MONIT/intelligence.go
```diff
@@ -0,0 +1,320 @@
...
+
+	ifAnyAlert := false
+
+	for _, each := range filteredAlerts.Data {
+		for k, v := range each.Labels {
+			if k == configJSON.SearchingLabel {
+				for _, ins := range strings.Split(v.(string), ",") {
+					if val, ok := silenceMap[ins]; ok {
+						ifAnyAlert = true
+						silence(val, each)
```
vkuznet, line 209: you should return error from silence and properly handle it here

src/go/MONIT/intelligence.go
```diff
@@ -0,0 +1,320 @@
...
+
+var configJSON config
+
+//-------STRUCTS---------
+
+//function for get request on /api/v1/alerts alertmanager endpoint for fetching alerts.
+func get(data interface{}) {
+
+	//GET API for fetching only GGUS alerts.
+	apiurl := configJSON.CMSMONURL + configJSON.GetAlertsAPI
```
vkuznet, line 89: replace with dedicated function to construct and validate urls

src/go/MONIT/intelligence.go
indrarahul2013: Valentin,
When we create a systemd service it will be helpful if we keep Environment variable. We can easily pass envs to sytemd service file like this.

`Environment="INTELLIGENCE_CONFIG_PATH"`

indrarahul2013:
@vkuznet I have added the configFile varible as flag. Though, I think we should keep Environment variable for systemd service. However, You can merge the PR, if it's alright.

src/go/MONIT/intelligence.go
vkuznet: we're fully migrating to k8s, therefore there is no need for systems as we'll most likely run it as service

indrarahul2013: Okay.

PR #66, opened 2020-07-04
updated intelligence.go added regex matching, configured seperator

indrarahul2013:
- added regex matching
- added both ["," , " "] seperator
- also configured seperator in case new seperator is introduced


indrarahul2013:
@vkuznet You can merge this PR

PR #67, opened 2020-07-07
Added Intelligence Module in src/go/

indrarahul2013:
Few Findings :-
- We will have to change the default "severity: info" to "severity:<others>" in case of SSB alerts because "info" is one of the value among <info,warning,medium,high,urgent> and we need to have unique pairs of alertname,severity to silence the alert. So there can be one case where "info" will be assigned and then on silencing the alert will get silenced even if it was not meant to be.
- GGUS works fine as default "severity:ticket" is assigned.


src/go/intelligence/pipeline/fetch_alert.go
```diff
@@ -0,0 +1,100 @@
...
+	err = json.Unmarshal(byteValue, &data)
+	if err != nil {
+		if utils.ConfigJSON.Verbose > 0 {
+			log.Println(string(byteValue))
+		}
+		log.Printf("Unable to parse JSON Data from AlertManager GET API, error: %v\n", err)
+		return data, err
+	}
+
+	if utils.ConfigJSON.Verbose > 1 {
```
vkuznet, line 90: This block should go above unmarshal the body since in debug mode we need to see response before we try to unmarshal it. For instance, if response fails to unmarshal this block will never be called.

src/go/intelligence/pipeline/keyword_matching.go
```diff
@@ -0,0 +1,90 @@
...
+// Created    : Wed, 1 July 2020 11:04:01 GMT
+// Description: CERN MONIT infrastructure Intelligence Module
+
+//KeywordMatching function finds defined keywords in the shortDescription of alerts and assign severity level accordingly
+func KeywordMatching(data <-chan models.AmJSON) <-chan models.AmJSON {
+
+	dataWithSeverity := make(chan models.AmJSON)
+	go func() {
+		for each := range data {
+			if each.Labels["service"] == "SSB" {
```
vkuznet, line 20: This should be generalized via local map and reflection.

src/go/intelligence/pipeline/silence_alert.go
```diff
@@ -0,0 +1,128 @@
...
+
+	for k, v := range data.Labels {
+		if k == utils.ConfigJSON.UniqueLabel {
+			if val, ok := v.(string); ok {
+				alertnameMatcher.Value = val
+			}
+		}
+
+		if k == utils.ConfigJSON.SeverityLabel {
+			if data.Labels["service"] == "SSB" {
```
vkuznet, line 64: This should be generalized via config map of severity levels.

src/go/intelligence/pipeline/preprocess.go
```diff
@@ -0,0 +1,36 @@
...
+		for each := range data {
+			for key, value := range each.Annotations {
+				if key == utils.ConfigJSON.SsbKeywordLabel {
+					if val, ok := value.(string); ok {
+						each.Annotations[key] = strings.ToLower(val)
+					}
+				}
+			}
+
+			if each.Labels["service"] == "SSB" || each.Labels["service"] == "GGUS" {
```
vkuznet, line 28: this should be generalized if you'll use config map for reflection and only loop over config map keys.

src/go/intelligence/pipeline/fetch_alert.go
indrarahul2013: Fixed this.

src/go/intelligence/pipeline/preprocess.go
indrarahul2013: Fixed this.

src/go/intelligence/pipeline/silence_alert.go
indrarahul2013: Fixed this.

src/go/intelligence/pipeline/keyword_matching.go
indrarahul2013: Added this.

indrarahul2013:
@vkuznet You can merge the PR now.

src/go/intelligence/pipeline/ml_box.go
```diff
@@ -0,0 +1,24 @@
...
+// Description: CERN MONIT infrastructure Intelligence Module
+
+//MlBox Machine Learning predicted Data
+//To be implemented
+func MlBox(data <-chan models.AmJSON) <-chan models.AmJSON {
+
+	predictedData := make(chan models.AmJSON)
+	go func() {
+		for d := range data {
+			predictedData <- d
```
vkuznet, line 19: Put doc string here: IMPLEMENT THE LOGIC OF ML PREDICTIONS, and write that so far we just return the data back without any logic

src/go/intelligence/main.go
```diff
@@ -0,0 +1,58 @@
...
+			pipeline.KeywordMatching(
+				pipeline.Preprocess(
+					pipeline.FetchAlert())))))
+
+	for d := range a {
+		processedData = append(processedData, d)
+	}
+
+	if utils.ConfigJSON.Server.Verbose > 2 {
+		fmt.Printf("Processed Alerts Data: %s\n", processedData)
```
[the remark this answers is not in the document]
indrarahul2013: Fixed.

src/go/intelligence/models/models.go
```diff
@@ -0,0 +1,79 @@
+package models
+
+import "time"
+
+// Module     : intelligence
+// Author     : Rahul Indra <indrarahul2013 AT gmail dot com>
+// Created    : Wed, 1 July 2020 11:04:01 GMT
+// Description: CERN MONIT infrastructure Intelligence Module
+
+//AmJSON AlertManager API acceptable JSON Data
```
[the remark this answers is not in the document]
indrarahul2013: Added doc string for each attributes.

src/go/intelligence/pipeline/keyword_matching.go
```diff
@@ -0,0 +1,103 @@
...
+	"reflect"
+	"strings"
+)
+
+// Module     : intelligence
+// Author     : Rahul Indra <indrarahul2013 AT gmail dot com>
+// Created    : Wed, 1 July 2020 11:04:01 GMT
+// Description: CERN MONIT infrastructure Intelligence Module
+
+//KeywordMatchFunction sd
```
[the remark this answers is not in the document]
indrarahul2013: Fixed this.

src/go/intelligence/pipeline/ml_box.go
indrarahul2013: Fixed This.

src/go/intelligence/pipeline/keyword_matching.go
```diff
@@ -0,0 +1,103 @@
...
+
+			dataWithSeverity <- each
+		}
+		close(dataWithSeverity)
+	}()
+
+	return dataWithSeverity
+}
+
+//SsbKeywordMatching SSB alerts KeywordMatching function
```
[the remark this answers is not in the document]
indrarahul2013: SSB Alerts and GGUS Alerts have similar logic for keyword matching and thus the function can be generalized and we can drop the use of reflection. Should I drop reflection logic ?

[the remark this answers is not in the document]
vkuznet: I showed you reflect functionality to generalize usage of multiple functions via their string names. But if your keyword matching function is unique there is no need for reflection logic. The generalization comes in terms of alert structures, and if it is applied to any alerts then we only need one keyword matching function.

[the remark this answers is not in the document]
indrarahul2013: Yes then I tried to make one function which works for all services and it works. Made a PR you can merge now.

indrarahul2013:
@vkuznet You can merge

indrarahul2013:
Valentin, 
I have to add one more feature to reduce the redundancy in Silence section, I see unnecessary copies of same silences. 
I will add this feature and make suggested changes and make a commit for the same. I understand README.md file should reside in src/go/intelligence/ ?

vkuznet:
Yes

indrarahul2013:
@vkuznet You can merge the PR now.

PR #70, opened 2020-07-11
Adding annotations to Dashboards from SSB Alerts/intelligence module

indrarahul2013:


src/go/intelligence/pipeline/add_annotations.go
```diff
@@ -0,0 +1,123 @@
...
+// Description: CMS MONIT infrastructure Intelligence Module
+
+//AddAnnotation function for adding annotations to dashboards
+func AddAnnotation(data <-chan models.AmJSON) <-chan models.AmJSON {
+
+	dataAfterAnnotation := make(chan models.AmJSON)
+
+	go func() {
+		dashboards := utils.FindDashboard()
+		for each := range data {
```
vkuznet, line 29: Please provide clear explanation of such complex structure, so far it is O(n^5) algorithm which by any definition is VERY slow. I think you should split it apart and provide O(n) algorithm. For instance, right series of for loops (rather netested one), where you need to find the following:
- find list of dashboards where annotations should be applied.
- find list of services in config
- find list of services in Annotation map
- take the intersection of aforementioned configs
- loop over final list of services and apply it to your list of dashboads.

Each loop is O(n), therefore the above logic will be N^4 times faster of your approach.

src/go/intelligence/pipeline/add_annotations.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		log.Printf("Unable to get response from %s, error: %s", apiURL, err)
+	}
+	if utils.ConfigJSON.Server.Verbose > 1 {
+		dump, err := httputil.DumpResponse(resp, true)
+		if err == nil {
+			log.Println("response:", string(dump))
+		}
+	}
+	body, _ := ioutil.ReadAll(resp.Body)
```
vkuznet, line 117: please close response body with defer function, i.e.
```
defer resp.Body.Close()
```

indrarahul2013: Fixed this.

src/go/intelligence/pipeline/add_annotations.go
indrarahul2013: @vkuznet  This is true. I have optimized the code. Please review

src/go/intelligence/pipeline/add_annotations.go
```diff
@@ -0,0 +1,139 @@
...
+// Created    : Wed, 1 July 2020 11:04:01 GMT
+// Description: CMS MONIT infrastructure Intelligence Module
+
+//AddAnnotation function for adding annotations to dashboards
+func AddAnnotation(data <-chan models.AmJSON) <-chan models.AmJSON {
+
+	dataAfterAnnotation := make(chan models.AmJSON)
+
+	go func() {
+		dashboards := utils.FindDashboard()
```
vkuznet, line 28: Here you don't need to finddashboard every single time since this information is static. Instead, you should add a cache and update the cache periodically, e.g. once a day. This will save time, traffic and number of calls to upstream server. For cache, please define a struct which will hold the dashboard map and expiration timestamp. Then add FindDatashboard method to the struct which will first check the cache, if it is expired it will use utils.FindDashboard to fetch new map, update the map, update the expiration time and return it.

indrarahul2013: @vkuznet This will decrease our time compexity for sure. I have added caching feature for the dashboards.

src/go/intelligence/utils/utils.go
```diff
@@ -16,10 +21,33 @@ import (
...
+
+//UpdateDashboardCache - function for updating the dashboards cache on expiration
+func (dCache *DashboardsCache) UpdateDashboardCache() {
+
+	if !FirstRunSinceRestart && !(dCache.Expiration.Equal(time.Now())) {
+		return
+	}
+
+	dCache.Dashboards = findDashboards()
+	dCache.Expiration = time.Now().Add(24 * ConfigJSON.AnnotationDashboard.DashboardsCacheExpiration * time.Hour)
```
vkuznet, line 44: I think you should remove 24 factor here.

src/go/intelligence/utils/utils.go
```diff
@@ -16,10 +21,33 @@ import (
...
+//DashboardsCache - a cache for storing dashboards and Expiration time for updating the cache
+type DashboardsCache struct {
+	Dashboards models.AllDashboardsFetched
+	Expiration time.Time
+}
+
+//UpdateDashboardCache - function for updating the dashboards cache on expiration
+func (dCache *DashboardsCache) UpdateDashboardCache() {
+
+	if !FirstRunSinceRestart && !(dCache.Expiration.Equal(time.Now())) {
```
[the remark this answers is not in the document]
indrarahul2013: Fixed the granularity & condition.

src/go/intelligence/utils/utils.go
indrarahul2013: Removed 24 factor. I was thinking we will have Expiration in terms of days, so 24.

PR #73, opened 2020-07-25
Race Condn elimination, Deletion of Silences, Proper handling of Suppressed Alerts

indrarahul2013:
Changes Made
- Race condition on Maps handled.
- Deletion of Expired Silences when original Alerts are resolved. 
- Proper handling of suppressed alerts.

### DIFF
--- a/src/go/intelligence/pipeline/delete_old_silents.go
+++ b/src/go/intelligence/pipeline/delete_old_silents.go
@@ -1,0 +1,112 @@
+package pipeline
+
+import (
+	"errors"
+	"go/intelligence/models"
+	"go/intelligence/utils"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"time"
+)
+
+// Module     : intelligence
+// Author     : Rahul Indra <indrarahul2013 AT gmail dot com>
+// Created    : Wed, 1 July 2020 11:04:01 GMT
+// Description: CMS MONIT infrastructure Intelligence Module
+
+//DeleteSilence - function for deleting expired silences
+func DeleteSilence(data <-chan models.AmJSON) <-chan models.AmJSON {
+
+	finalData := make(chan models.AmJSON)
+	go func() {
+		deleteSilenceHelper()
+		for each := range data {
+			finalData <- each
+		}
+		close(finalData)
+	}()
+	return finalData
+}
+
+//deleteSilenceHelper - helper function for deleting a silence
+func deleteSilenceHelper() {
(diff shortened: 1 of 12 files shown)


src/go/intelligence/pipeline/delete_old_silents.go
```diff
@@ -0,0 +1,112 @@
...
+//DeleteSilence - function for deleting expired silences
+func DeleteSilence(data <-chan models.AmJSON) <-chan models.AmJSON {
+
+	finalData := make(chan models.AmJSON)
+	go func() {
+		deleteSilenceHelper()
+		for each := range data {
+			finalData <- each
+		}
+		close(finalData)
```
vkuznet, line 27: this channel may not be closed if above code fails and exit prematurely. It is much safer to use `defer close(finalData)` at the beginning of this function. The defer call will ensure that channel is closed regardless of body function execution.

src/go/intelligence/pipeline/fetch_alert.go
```diff
@@ -21,7 +15,12 @@ import (
 func FetchAlert() <-chan models.AmJSON {
 	fetchedData := make(chan models.AmJSON)
 	go func() {
```
vkuznet, line 17: the same problem with fetchedData channel, it may never be closed if the body of this function fails. Please use `defer close(fetchedData)` at the beginning of this goroutine.

src/go/intelligence/pipeline/keyword_matching.go
```diff
@@ -18,12 +18,13 @@ func KeywordMatching(data <-chan models.AmJSON) <-chan models.AmJSON {
...
-				if each.Labels[utils.ConfigJSON.Alerts.ServiceLabel] == service.Name {
-					keywordMatchingHelper(&each, service)
+				if changedData.Labels[utils.ConfigJSON.Alerts.ServiceLabel] == service.Name {
+					keywordMatchingHelper(&changedData, service)
 				}
 			}
-			dataWithSeverity <- each
+			dataWithSeverity <- changedData
 		}
 		close(dataWithSeverity)
```
vkuznet, line 29: again the same problem with closing channel to late. Please fix all close statements with `defer` and move them up in a goroutine code.

src/go/intelligence/pipeline/delete_old_silents.go
indrarahul2013: Resolved.

indrarahul2013:
@vkuznet Now you can merge the PR.

PR #76, opened 2020-07-30
intelligent module/ -dryRun, -iter flag, before/after logging, added + minor changes

indrarahul2013:


src/go/intelligence/main.go
```diff
@@ -14,6 +14,9 @@ import (
 // Created    : Wed, 1 July 2020 11:04:01 GMT
 // Description: CMS MONIT infrastructure Intelligence Module
 
+//variable storing custom defined no. of iteration
+var iter int
```
vkuznet, line 18: why this is a global variable?

src/go/intelligence/main.go
```diff
@@ -36,12 +39,55 @@ func run() {
...
+		}
+		log.Printf("No of Active Silences in AlertManager after runnint int module : %d\n", utils.NoOfActiveSilencesAfterIntModule)
+		log.Printf("No of Expired Silences in AlertManager after runnint int module : %d\n", utils.NoOfExpiredSilencesAfterIntModule)
+		log.Printf("No of Pending Silences in AlertManager after runnint int module : %d\n", utils.NoOfPendingSilencesAfterIntModule)
+	} else {
+		log.Printf("Dry Run.. No of Alerts meant to be pushed into AlertManager : %d\n", utils.NoOfPushedAlerts)
+		log.Printf("Dry Run.. No of new Silences meant to be created into AlertManager : %d\n", utils.NoOfNewSilencedAlerts)
+	}
+
+	time.Sleep(utils.ConfigJSON.Server.Interval * time.Second)
```
vkuznet, line 79: it does not belong here since the entire function does not have an infinite loop and neither provide daemon functionality

src/go/intelligence/main.go
```diff
@@ -36,12 +39,55 @@ func run() {
 	}
 }
 
+func runNoOfTimes() {
+	utils.NoOfAlertsBeforeIntModule = 0
```
vkuznet, line 43: all of these counters are not atomic and therefore can be screwed in concurrent calls

src/go/intelligence/main.go
```diff
@@ -36,12 +39,55 @@ func run() {
 	}
 }
 
+func runNoOfTimes() {
```
vkuznet, line 42: Please use full names without shortcuts

src/go/intelligence/main.go
```diff
@@ -36,12 +39,55 @@ func run() {
...
+	}
+}
+
 func runInfinite() {
-	utils.FirstRunSinceRestart = true
 	for true {
-		run()
-		utils.FirstRunSinceRestart = false
-		time.Sleep(utils.ConfigJSON.Server.Interval * time.Second)
+		runNoOfTimes()
```
vkuznet, line 90: sleep function should go after runNoOfTimes

src/go/intelligence/utils/utils.go
```diff
@@ -201,16 +239,95 @@ func findDashboards() models.AllDashboardsFetched {
...
+	err = json.Unmarshal(byteValue, &data)
+	if err != nil {
+		if ConfigJSON.Server.Verbose > 0 {
+			log.Println(string(byteValue))
+		}
+		log.Printf("Unable to parse JSON Data from AlertManager Silence GET API, error: %v\n", err)
+		return data, 0, 0, 0, err
+	}
+
+	expiredLength := 0
```
vkuznet, line 302: I don't really see needs for making new variables and complicating function signature. Why not to put all counters into data structure itself?

src/go/intelligence/main.go
indrarahul2013: Fixed

src/go/intelligence/main.go
indrarahul2013: Changed the name to changesLogging() containing only logging logics.

src/go/intelligence/main.go
indrarahul2013: Removed it.

indrarahul2013:
@vkuznet Getting zero counter value at the end of the pipeline won't happen. This is because when we fetch current alerts in the AM, some of them will have changed values and some of them are newly created. Those which have changed value will not be processed again and thus the difference between noOfAlerts counter after and before the int module won't be always zero. Similarly, the silences Counter goes. I propose we should fetch all the alert/silences data throughout the pipeline and we get the following values:-
- No of Current Alerts in AM (While Fetch Alerts)
- No of Current Active/Expired/Pending Silences in AM (while Preprocess)
- No of newly Pushed Alerts (while Push Alert)
- No of newly created Silence ( while Silence Alert)

After completion of one iteration. We again ask AM about Current Alerts & Current Active/Expired/Pending Silences and print it in the log. If we see differences then we see there is some problem.

src/go/intelligence/main.go
indrarahul2013: @vkuznet I have changed the counters into one data structure and they are being incremented/decremented in the pipeline execution only. I think concurrent calls issue can happen only in the pipeline. But, here we are making changes in the infinite loop and after each loop now I am resetting the counter data structure.

src/go/intelligence/utils/utils.go
indrarahul2013: Fixed this issue.

src/go/intelligence/pipeline/fetch_alert.go
```diff
@@ -14,17 +14,20 @@ import (
 //FetchAlert - function for fetching all active alerts from AlertManager
 func FetchAlert() <-chan models.AmJSON {
 	fetchedData := make(chan models.AmJSON)
+
```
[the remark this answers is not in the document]
indrarahul2013: There's one specific reason for pulling this block out of goRoutine. I want to first fetch all alerts and then start the concurrent calls. This is because there was a discrepancy with matching with Alerts going to be processed and already silenced alerts. So to make sure we all have alert data before calling goroutines.

indrarahul2013:
@vkuznet We were sending single alert data in the pipeline before on the channel and thus it was listening like:- 

`for each := range data {}`

Now if we add alerts, silences on pipeline we won't be able to pass single alert and silence because there will be no relation between those single alert and silences (i.e. they will be completely different, one can be new alert in AM and silence can be older alert which has been silenced before).

So we can push (alert, All Silences, counters) on the pipeline. (i.e. now channel will get alert one by one but it will also contain allSilences fetched and allCounters)

vkuznet:
ok, then it will be redundant to have a single alert and all silences. Then I think you should get all silences before the alerts pipeine. Doing this way all silances will be setup and you'll only update it as you'll walk through alerts. Now you have updateSilencedMap within alert preprocess step which is very confusing to me.

Please write in plain English the logic of the entire pipeline and all steps you need to do that we can agree on it.

indrarahul2013:
I will try to put all details here including your suggestion:-
1) the main func will run all pipeline logic.
2) pipeline starts with fetching alerts, 
3) then to process each alert seeing if it has already been processed and silenced if they are then ignoring that alert, otherwise pass it to the next pipeline component. ( So here we require the SilencedMap which stores all those alerts which are in Silence Mode which indicates they are processed and we should not repeat the intelligence process again for them )
4) Then the alert comes to keyword matching were keywords are matched and accordingly severity is assigned.
5) Passes through ML Box doing nothing.
6) then the processed alert is pushed to AM and after this,
7) Then the old alert with default severity is silenced
8) Then Some resolved alerts that are silenced (i.e. when GGUS alerts are resolved) are deleted.

So in the SilencedMap being updated in step 3 (according to your suggestion) can be pushed to the main function.

Regarding Counters:- 
When we are fetching the alerts at step 2, we will count the number of alerts in the AM (i.e. before intelligence module does its stuff)
Then when we go to preprocessing step 3, we will count all active, expired silences when we update our SilenceMap
When we push Alerts we count how many alerts got pushed. 
When we create new Silences we will add 1 to the Active Silence counter which we modified at step 3
When we delete a resolved alert's Silence we will add 1 the Expired Silence counter which we modified at step 3.

Now at the end of pipeline. We will end up having following counters:-
- No Of Alerts
- No Of Active Silences
- No Of Expired Silences
- No of Alerts Pushed

Now to verify if everything went well at the end of pipeline. We will again fetch Alerts/Silences from AM and count them and check if they matches with the counters above. If they match go to next iteration otherwise stop the module

vkuznet:
Rahul, could you please these details into code itself that we can easily find them.
How the tests are doing? And, does this PR is ready for merge or you're planning to add something else?

indrarahul2013:
@vkuznet I have added the test script and the above commit is for the same.

vkuznet:
yes, add to the code the logic if the code in plain English as it is shown in this ticket.

vkuznet:
Rahul, I asked you to add the notes to the codebase, then I can merge. Please do so.

indrarahul2013:
@vkuznet now you can merge this PR.

PR #95, opened 2020-08-24
Int Module - Addition of multiple dashboard annotation based on specific set of keywords

indrarahul2013:
Changes :-

- Change in the config file which now supports annotations based on a different set of keywords to dashboards with a different tag. The list of annotations can be easily configured with a specific set of keywords for specific dashboards identified by tags.
```
"annotationMap": {
        "label": "shortDescription",
        "annotations": [
          {
            "actions": ["intervention", "outage"],
            "systems": ["network", "database", "db"],
            "tags": ["cmsweb-play"]
          },

          {
            "actions": ["update", "upgrade"],
            "systems": ["network", "database", "db"],
            "tags": ["cmsweb-play2"]
          }
        ],
        "urlLabel": "URL"
      }
```
- Change in the add_annotation component of the pipeline, where the intersection between the dashboards is found and accordingly they are annotated.

### DIFF
--- a/src/go/intelligence/int_test/test_config.json
+++ b/src/go/intelligence/int_test/test_config.json
@@ -22,7 +22,7 @@
     "url" : "https://monit-grafana.cern.ch",
     "dashboardSearchAPI" : "/api/search",
     "annotationAPI" : "/api/annotations",
-    "tags" : ["cmsweb-play"],
+    "tags" : ["cmsweb-play","cmsweb-play2"],
     "token" : "<YOUR-GRAFANA-PROXY-TOKEN>",
     "dashboardsCacheExpiration" : 1 
   },
@@ -66,8 +66,19 @@
       },
       "annotationMap": {
         "label": "shortDescription",
-        "actions": ["intervention", "outage"],
-        "systems": ["network", "database", "db"],
+        "annotations": [
+          {
+            "actions": ["intervention", "outage"],
+            "systems": ["network", "database", "db"],
+            "tags": ["cmsweb-play"]
+          },
+
+          {
+            "actions": ["update", "upgrade"],
+            "systems": ["network", "database", "db"],
+            "tags": ["cmsweb-play2"]
+          }
+        ],
         "urlLabel": "URL"
       }
     },
@@ -83,8 +94,13 @@
       },
       "annotationMap": {
         "label": "Subject",
-        "actions": ["transfers", "outage"],
-        "systems": ["rucio"],
+        "annotations": [
+          {
+            "actions": ["transfer", "outage"],
+            "systems": ["rucio"],
+            "tags": ["cmsweb-play2"]
+          }
+        ],
         "urlLabel": "URL"
       }
     }
--- a/src/go/intelligence/models/models.go
+++ b/src/go/intelligence/models/models.go
@@ -48,7 +48,7 @@ type AllSilences struct {
 }
 
 //AllDashboardsFetched is an array of all Dashboards' information having given tags in common
-type AllDashboardsFetched []struct {
+type AllDashboardsFetched struct {
 	ID          float64  `json:"id"`
 	UID         string   `json:"uid"`
 	Title       string   `json:"title"`
@@ -121,11 +121,17 @@ type silence struct {
 	SilenceStatus []string `json:"silenceStatus"` //Labels for status of the silence
 }
 
+//annotations struct for storing specific set of keywords & set of dashboards to annotate when these keywords are matched in alerts.
+type annotationsData struct {
+	Actions []string `json:"actions"`
+	Systems []string `json:"systems"`
+	Tags    []string `json:"tags"`
+}
+
 type annotationMap struct {
-	Label    string   `json:"label"`
-	Actions  []string `json:"actions"`
-	Systems  []string `json:"systems"`
-	URLLabel string   `json:"urlLabel"`
+	Label           string            `json:"label"`
+	AnnotationsData []annotationsData `json:"annotations"`
+	URLLabel        string            `json:"urlLabel"`
 }
 
 //Service data struct
--- a/src/go/intelligence/models/models.go
+++ b/src/go/intelligence/models/models.go
@@ -49,28 +49,28 @@ type AllSilences struct {
 
 //AllDashboardsFetched is an array of all Dashboards' information having given tags in common
 type AllDashboardsFetched struct {
-	ID          float64  `json:"id"`
-	UID         string   `json:"uid"`
-	Title       string   `json:"title"`
-	URI         string   `json:"uri"`
-	URL         string   `json:"url"`
-	Slug        string   `json:"slug"`
-	Type        string   `json:"type"`
-	Tags        []string `json:"tags"`
-	IsStarred   bool     `json:"isStarred"`
-	FolderID    float64  `json:"folderId"`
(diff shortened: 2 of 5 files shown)


src/go/intelligence/models/models.go
```diff
@@ -121,11 +121,17 @@ type silence struct {
 	SilenceStatus []string `json:"silenceStatus"` //Labels for status of the silence
 }
 
+//annotations struct for storing specific set of keywords & set of dashboards to annotate when these keywords are matched in alerts.
+type annotationsData struct {
+	Actions []string `json:"actions"`
```
vkuznet, line 126: please add documentation strings for EACH member of the struct

src/go/intelligence/models/models.go
```diff
@@ -121,11 +121,17 @@ type silence struct {
...
+	Systems []string `json:"systems"`
+	Tags    []string `json:"tags"`
+}
+
 type annotationMap struct {
-	Label    string   `json:"label"`
-	Actions  []string `json:"actions"`
-	Systems  []string `json:"systems"`
-	URLLabel string   `json:"urlLabel"`
+	Label           string            `json:"label"`
```
vkuznet, line 132: add doc strings for each member of the struct

src/go/intelligence/int_test/test_config.json
```diff
@@ -66,8 +66,19 @@
...
+            "tags": ["cmsweb-play"]
+          },
+
+          {
+            "actions": ["update", "upgrade"],
+            "systems": ["network", "database", "db"],
+            "tags": ["cmsweb-play2"]
+          }
+        ],
         "urlLabel": "URL"
```
vkuznet, line 82: I think urlLabel will be part of each annotation since it may refer to different URLs, e.g. one for SSB and another for GGUS. Moreover. do you really need to call it urlLabel, it represents the URL and I think it should be called URL in the configuration.

src/go/intelligence/int_test/test_config.json
```diff
@@ -66,8 +66,19 @@
       },
       "annotationMap": {
         "label": "shortDescription",
```
vkuznet, line 68: Does the label should be applied to all annotations or individual annotations may have individual labels?

src/go/intelligence/pipeline/add_annotations.go
```diff
@@ -84,6 +141,26 @@ func makeHTMLhref(url string) string {
 	return "<a target=_blank href=" + url + ">URL</a>"
 }
 
+//getTagsIntersection for finding intersection between two list of dashboard tags.
```
vkuznet, line 144: the description and name of this function is misleading. It does not find the intersection, since the intersection is a list of something, instead what it does it finds if entries from one list appear in another list. If so, the implementation is not correct either since you can simply loop over one list and find if its entry exists in another, if so you can return immediately a true boolean.

Please clarify the function behavior, rename it properly and simply.

src/go/intelligence/int_test/test_config.json
indrarahul2013: Valentin, yes we need to call it urlLabel because in alert data it's a field "URL", so if there's change in the alerting services in future in case, we can change the urlLabel value here accordingly. Moreover, each service has unique urlLabel identifier, that said we don't need to add multiple times in the annotations array above. The above example is of one service, if you check the test_config.json you can see I have maintained different urlLabel for different services.

src/go/intelligence/int_test/test_config.json
indrarahul2013: here the label selectively chooses field in the alert data where all information about the alert is defined by the operators. Then predefined keywords are matched in this descriptive field. So I think it is not necessary to put it into individual annotations. Because operators use one field to put details in (eg. for SSB its "shortDescription, for GGUS its "Subject")

src/go/intelligence/pipeline/add_annotations.go
indrarahul2013: Valentin,
It is actually finding the intersection between two lists, but what I am returning is if the resultant list after getting an intersection is empty or not. So, intuitively I am checking if there are any dashboards common between all dashboards and dashboards specific to an annotation.

```

//getTagsIntersection for finding intersection between two list of dashboard tags.
func getTagsIntersection(annotationDashboardTags, allDashboardsTags []string) bool {

	var intersectionResult []string      ------>>>> HERE THE INTERSECTION RESULT IS GETTING STORED

	hashMap := make(map[string]bool) //a hashmap which helps to reduce the intersection operation from O(n^2) to O(n).

	for _, val := range allDashboardsTags {
		hashMap[val] = true
	}

	for _, each := range annotationDashboardTags {
		if _, ok := hashMap[each]; ok {
			intersectionResult = append(intersectionResult, each)
		}
	}

	return len(intersectionResult) > 0            ----> CHECKING IF THE RESULTANT LIST AFTER INTERSECTION IS NOT EMPTY
}
```

indrarahul2013: I used intersection concept using hashmap to optimize the process from O(n^2) to O(n)

src/go/intelligence/models/models.go
indrarahul2013: added.

src/go/intelligence/models/models.go
indrarahul2013: added

src/go/intelligence/pipeline/add_annotations.go
indrarahul2013: I have made a few changes in description now.

src/go/intelligence/utils/utils.go
```diff
@@ -54,7 +54,7 @@ var SuppressedAlertsDataReadWriteLock sync.RWMutex
 
 //DashboardsCache - a cache for storing dashboards and Expiration time for updating the cache
 type DashboardsCache struct {
-	Dashboards models.AllDashboardsFetched
+	Dashboards map[float64]models.AllDashboardsFetched
```
vkuznet, line 57: Why map keys are float64? As far as I see from the code those are simple unsigned ints. Please double check and fix appropriately.

indrarahul2013: This is because Grafana dashboard search API returns the value in float64. Here you can see the model for the dashboard struct.

```
type AllDashboardsFetched struct {
	ID          float64  `json:"id"`          //ID for the dashboard
	UID         string   `json:"uid"`         //UID for the dashboard
	Title       string   `json:"title"`       //Title for the dashboard
	URI         string   `json:"uri"`         //URI of the dashboard
	URL         string   `json:"url"`         //URL of the dashboard
	Slug        string   `json:"slug"`        //Slug of the dashboard
	Type        string   `json:"type"`        //Type of the dashboard
	Tags        []string `json:"tags"`        //All tags for the dashboard	(eg. prod, jobs, cmsweb etc.)
	IsStarred   bool     `json:"isStarred"`   //if dashboard is starred
	FolderID    float64  `json:"folderId"`    //ID for the folder
	FolderUID   string   `json:"folderUid"`   //UID for the folder
	FolderTitle string   `json:"folderTitle"` //Title of the folder
	FolderURL   string   `json:"folderUrl"`   //URL of the folder
}
```

vkuznet: ok, thanks for checking (I was expecting something like that).

PR #97, opened 2020-09-01
annotationManager added

indrarahul2013:
- Lists all annotations.
- Lists all annotations with specified tags.
- Lists all annotations over some time period.
- Deletes listed annotations.

```
Usage: annotationManager [options]
  -action string
    	Action to be performed. [list, create, delete, deleteall, update] Default: list
  -annotation string
    	Annotation text
  -annotationID int
    	Annotation ID required in making delete api call for deleting an annotation
  -generateConfig
    	Flag for generating default config
  -tags string
    	List of tags seperated by comma
  -token string
    	Authentication token to use (Optional-can be stored in config file)
  -trange string
    	Time Range for filtering annotations.
  -verbose int
    	verbosity level, can be overwritten in config (Optional-can be stored in config file)

Environments:
	CONFIG_PATH:	 Config to use, default ($HOME/.annotationManagerConfig.json)

Examples:
	ALTHOUGH ACTION SHOULD BE (-action=list) FOR GETTING ALL ANNOTATIONS BUT ACTION FLAG CAN BE DROPPED AS LIST IS DEFAULT ACTION!

	Get all annotations with specified tags:
	    annotationManager -action=list -tags=das,cmsmon-int

	Use (d/day/days) for days, (h/hr/hrs) for hours, (m/min/mins) for minutes and (s/sec/secs) for seconds.

	Get all annotations with specified tags on time period range (from last 7days to now)
	    annotationManager -action=list -tags=das,cmsmon-int -trange=7d-now

	Get all annotations with specified tags on time period range (from last 14days to last 7days)
	    annotationManager -action=list -tags=das,cmsmon-int -trange=14d-7d

	Get all annotations with specified tags on time period range (using dates seperated by '--')
	    annotationManager -action=list -tags=das,cmsmon-int -trange=2020-09-01T19:49:50.206Z--2020-09-01T19:49:50.206Z

	Delete all annotations once you find appropriate annotations to delete from the above commands.
	    annotationManager -action=deleteall -tags=das,cmsmon-int -trange=7d-now
	    annotationManager -action=deleteall -tags=das,cmsmon-int -trange=14d-7d
	    annotationManager -action=deleteall -tags=das,cmsmon-int -trange=2020-09-01T19:49:50.206Z--2020-09-01T19:49:50.206Z

	Delete an annotation with specific annotation id (/api/annotations/:id)
	    annotationManager -action=delete -annotationID=761323

	Create an annotation on dashboards with specific tags with time ranging from now to 2m
	    annotationManager -action=create -annotation="some message" -tags=das,cmsweb -trange=now-2m

	Update an annotation with specific annotation id with new annotation and tags. (/api/annotations/:id)
	(TAGS AND ANNOTATION ARE REQUIRED FLAGS OR OLD DATA WILL BE REPLACED WITH NULL VALUES. If you don't want to change annotation or tag, please send the old values.)
	    annotationManager -action=update -annotationID=761323 -annotation="some new message" -tags=newTag1,newTag2

	Update an annotation with specific annotation id with new start and end Time. (/api/annotations/:id)
	(TAGS AND ANNOTATION ARE REQUIRED FLAGS OR OLD DATA WILL BE REPLACED WITH NULL VALUES. If you don't want to change annotation or tag, please send the old values.)
	    annotationManager -action=update -annotationID=761323 -annotation="old message" -tags=oldTag1,oldTag2 -trange=now-5m

```

### DIFF
--- a/src/go/MONIT/annotationManager.go
+++ b/src/go/MONIT/annotationManager.go
@@ -8,7 +8,6 @@ package main
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -91,7 +90,7 @@ type config struct {
 // -------STRUCTS---------
 
 // function for get request grafana annotations endpoint for fetching annotations.
-func getAnnotations(data *[]annotationData, tags []string) error {
+func getAnnotations(data *[]annotationData, tags []string) {
 
 	var headers [][]string
 	bearer := fmt.Sprintf("Bearer %s", configJSON.Token)
@@ -125,8 +124,7 @@ func getAnnotations(data *[]annotationData, tags []string) error {
 	}
 	req, err := http.NewRequest("GET", apiURL, nil)
 	if err != nil {
-		log.Printf("Unable to make request to %s, error: %s", apiURL, err)
-		return err
+		log.Fatalf("Unable to make request to %s, error: %s", apiURL, err)
 	}
 	for _, v := range headers {
 		if len(v) == 2 {
@@ -147,15 +145,14 @@ func getAnnotations(data *[]annotationData, tags []string) error {
 
 	resp, err := client.Do(req)
 	if err != nil {
-		return err
+		log.Fatalf("Response Error, error: %v\n", err)
 	}
 	defer resp.Body.Close()
 
 	byteValue, err := ioutil.ReadAll(resp.Body)
 
 	if err != nil {
-		log.Printf("Unable to read JSON Data from Grafana Annotation GET API, error: %v\n", err)
-		return err
+		log.Fatalf("Unable to read JSON Data from Grafana Annotation GET API, error: %v\n", err)
 	}
 
 	err = json.Unmarshal(byteValue, &data)
@@ -164,7 +161,6 @@ func getAnnotations(data *[]annotationData, tags []string) error {
 			log.Println(string(byteValue))
 		}
 		log.Fatalf("Unable to parse JSON Data from Grafana Annotation GET API, error: %v\n", err)
-		return err
 	}
 
 	if configJSON.Verbose > 1 {
@@ -173,19 +169,16 @@ func getAnnotations(data *[]annotationData, tags []string) error {
 			log.Println("Response: ", string(dump))
 		}
 	}
-
-	return nil
 }
 
 // function for delete request on grafana annotations endpoint for deleting an annotation.
-func deleteAnnotationHelper(annotationID int) error {
+func deleteAnnotationHelper(annotationID int) {
 
 	apiurl := fmt.Sprintf("%s%s%d", configJSON.GrafanaBaseURL, configJSON.DeleteAnnotationAPI, annotationID)
 
 	req, err := http.NewRequest("DELETE", apiurl, nil)
 	if err != nil {
-		log.Printf("Request Error, error: %v\n", err)
-		return err
+		log.Fatalf("Request Error, error: %v\n", err)
 	}
 
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", configJSON.Token))
@@ -204,12 +197,10 @@ func deleteAnnotationHelper(annotationID int) error {
 
 	resp, err := client.Do(req)
 	if err != nil {
-		log.Printf("Response Error, error: %v\n", err)
-		return err
+		log.Fatalf("Response Error, error: %v\n", err)
 	}
 	if resp.StatusCode != http.StatusOK {
-		log.Printf("Http Response Code Error, status code: %d", resp.StatusCode)
-		return errors.New("Http Response Code Error")
+		log.Fatalf("Http Response Code Error, status code: %d", resp.StatusCode)
 	}
 
 	defer resp.Body.Close()
@@ -220,7 +211,6 @@ func deleteAnnotationHelper(annotationID int) error {
 			log.Println("Response: ", string(dump))
 		}
 	}
-	return nil
 }
 
 //
@@ -374,10 +364,7 @@ func tabulate(data []annotationData) {
 func listAnnotation() []annotationData {
 	var aData []annotationData
 	tagList := parseTags(tags)
-	err := getAnnotations(&aData, tagList)
-	if err != nil {
-		log.Fatalf("Unable to fetch annotations, error :%v", err)
-	}
+	getAnnotations(&aData, tagList)
 
 	if len(aData) == 0 {
 		log.Printf("NO ANNOTATION FOUND\n")
@@ -644,12 +631,8 @@ func updateAnnotation() {
 
 // function which contains the logic of deleting an annotation
 func deleteOneAnnotation() {
-	err := deleteAnnotationHelper(annotationID)
-	if err != nil {
-		log.Fatalf("Unable to delete the annotation with ID %d, error :%v", annotationID, err)
-	} else {
-		log.Printf("Annotation with id: %d has been deleted successfully !", annotationID)
-	}
+	deleteAnnotationHelper(annotationID)
+	log.Printf("Annotation with id: %d has been deleted successfully !\n", annotationID)
 }
 
 // function which contains the logic of deleting multiple annotations
@@ -659,13 +642,7 @@ func deleteAllAnnotations() {
 	if aData != nil {
 		log.Printf("DELETING ALL OF THE ALERTS SHOWN ABOVE!!\n")
 		for _, each := range aData {
-			err := deleteAnnotationHelper(each.ID)
-			if err != nil {
-				if configJSON.Verbose > 1 {
-					log.Printf("Annotation Data: %v", each)
-				}
-				log.Fatalf("Unable to delete the annotation, error :%v", err)
-			}
+			deleteAnnotationHelper(each.ID)
 			if configJSON.Verbose > 2 {
 				log.Printf("Annotation Deleted for:\n%+v\n", each)
 			}
@@ -673,7 +650,7 @@ func deleteAllAnnotations() {
 		}
 		log.Printf("Successfully Deleted %d annotations!!\n", noOfDeletedAnnotation)
 	} else {
-		log.Printf("Unable to delete with null data!\n")
+		log.Fatalf("Unable to delete with null data!\n")
 	}
 }
 
@@ -812,12 +789,12 @@ func parseTimes(ifCreatingAnnotation bool) []int64 {
 
 		startTime, err := time.Parse(time.RFC3339, tms[0])
 		if err != nil {
-			log.Printf("Unable to parse time, error: %s", err)
+			log.Fatalf("Unable to parse time, error: %s", err)
 		}
 
 		endTime, err := time.Parse(time.RFC3339, tms[1])
 		if err != nil {
-			log.Printf("Unable to parse time, error: %s", err)
+			log.Fatalf("Unable to parse time, error: %s", err)
 		}
 
 		times = append(times, startTime.Unix()*int64(milliSec))
--- a/src/go/MONIT/annotationManager.go
+++ b/src/go/MONIT/annotationManager.go
@@ -1,0 +1,595 @@
+package main
+
+// File       : alert.go
+// Author     : Rahul Indra <indrarahul2013 AT gmail dot com>
+// Created    : Tue, 1 Sep 2020 14:08:00 GMT
+// Description: CERN MONIT infrastructure Alert CLI Tool
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,595 @@
+package main
+
+// File       : alert.go
```
vkuznet, line 3: this file is not alert.go, please fix

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,595 @@
...
+	Updated     int64    `json:"updated"`     //Timestamp when the annotation last updated
+	Time        int64    `json:"time"`        //Start Time of the annotation
+	TimeEnd     int64    `json:"timeEnd"`     //End Time of the annotation
+	Text        string   `json:"text"`        //Text field of the annotation
+	Tags        []string `json:"tags"`        //List of tags attached with the annotation
+}
+
+type config struct {
+	GrafanaBaseURL      string   `json:"grafanaBaseURL"`      // Grafana Annotation Base URL
+	GetAnnotationsAPI   string   `json:"getAnnotationsAPI"`   //API endpoint for fetching all annotations
```
vkuznet, line 70: be consistent with commend, either use `// comment` or `//comment`, I prefer extra space behind two slashes.

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,595 @@
...
+		times = append(times, t*milliSec) // times should be in milliseconds of Unix since epoch
+	}
+	return times
+}
+
+func main() {
+
+	flag.StringVar(&token, "token", "", "Authentication token to use")
+	flag.StringVar(&tags, "tags", "", "List of tags seperated by comma")
+	flag.StringVar(&action, "action", "", "Action to be performed, eg. delete, update etc.")
```
vkuznet, line 562: do not put **eg**, be precise which actions are support. I suggest that the tool should have this list: create, update, delete, list and you should implement them. And, clearly say which is default one (which is list as far as I can tell).

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,595 @@
...
+		}
+	}
+}
+
+//Function running all logics
+func run() {
+
+	var aData []annotationData
+	tagList := parseTags(tags)
+	err := getAnnotations(&aData, tagList)
```
vkuznet, line 372: I think this should go under default case in switch clause below

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,595 @@
...
+	if err != nil {
+		log.Fatalf("Unable to fetch annotations, error :%v", err)
+	}
+	tabulate(aData)
+
+	switch action {
```
vkuznet, line 378: the action switch should provide all cases: list, create, update, delete. So far I only see list (which is getAnnotations) and delete. Can you implement create and update?

indrarahul2013: Valentin as far as I see, monit tool already has the feature to create an annotation, and also our intelligence module is creating automatically. Do we also need to add it to this tool? 
Yes, update can be implemented.

src/go/MONIT/annotationManager.go
indrarahul2013: fixed

src/go/MONIT/annotationManager.go
indrarahul2013: fixed

src/go/MONIT/annotationManager.go
indrarahul2013: fixed

src/go/MONIT/annotationManager.go
indrarahul2013: We would require data for deleting annotations. So I have moved the tabulate() function to the default one. Which basically prints all found annotation in tabular form on stdout.

src/go/MONIT/annotationManager.go
vkuznet: So, it seems to me that we should review both annotation and alert tools to decide. So far I want to complete this one to cover all actions we need, then we can decide which functionality should be kept where. The alert tool should probably only handle alerts, while annotations tool should handle only annotations.

indrarahul2013: Done !

indrarahul2013:
@vkuznet You can merge the PR now.

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+			log.Println("response:", string(dump))
+		}
+	}
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("HTTP Respose error, code: %d", resp.StatusCode)
+	}
+
+	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(resp.Body)
```
vkuznet, line 597: could you please handle the error properly here and check the code that all errors will be logged

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+		log.Fatalf("Unable to get response from %s, error: %s", rurl, err)
+	}
+	if configJSON.Verbose > 1 {
+		dump, err := httputil.DumpResponse(resp, true)
+		if err == nil {
+			log.Println("response:", string(dump))
+		}
+	}
+	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(resp.Body)
```
vkuznet, line 496: error here should be properly handled

indrarahul2013: fixed !

src/go/MONIT/annotationManager.go
indrarahul2013: Fixed !

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,931 @@
...
+				rec["tags"] = parseTags(tags)
+				rec["text"] = annotation
+				data, err := json.Marshal(rec)
+				if err != nil {
+					log.Fatalf("Unable to marshal the data %+v, error %v\n", rec, err)
+				}
+				if configJSON.Verbose > 0 {
+					log.Printf("Add annotation: %+v", rec)
+				}
+				addAnnotation(data)
```
vkuznet, line 537: there is no check for error here

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,931 @@
...
+		updatedData["tags"] = parseTags(tags)
+	}
+
+	data, err := json.Marshal(updatedData)
+	if err != nil {
+		log.Fatalf("Unable to parse Data for Updation, update failed !, error: %v", err)
+	}
+
+	err = updateAnnotationHelper(annotationID, data)
+	if err != nil {
```
vkuznet, line 642: there is no need for error checking here if you'll move this block into updateAnnoataionHelper itself.

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,931 @@
...
+	if configJSON.Verbose > 1 {
+		dump, err := httputil.DumpResponse(resp, true)
+		if err == nil {
+			log.Println("response:", string(dump))
+		}
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Printf("Unable to read JSON Data from Grafana Annotation POST API, error: %v\n", err)
```
vkuznet, line 498: you should use log.Fatal here

vkuznet (changes requested):
Rahul, please review all error handling. You're doing double job of catching in error and handling it in upstream code (see my comments for examples). Instead, you should immediately use log.Fatal to force code to exit once you check the error at low level functions like ioutil calls.

src/go/MONIT/annotationManager.go
indrarahul2013: Fixed

src/go/MONIT/annotationManager.go
indrarahul2013: fixed

src/go/MONIT/annotationManager.go
indrarahul2013: it would be similar case as updateAnnotationHelper, there are different error checks inside the addAnnotation block.

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	byteValue, err := ioutil.ReadAll(resp.Body)
+
+	if err != nil {
+		log.Printf("Unable to read JSON Data from Grafana Annotation GET API, error: %v\n", err)
```
vkuznet, line 157: replace with Fatal and adjust code which uses function. It seems to me there is no need to return an error if all errors will be handled within this function body.

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+
+// function for delete request on grafana annotations endpoint for deleting an annotation.
+func deleteAnnotationHelper(annotationID int) error {
+
+	apiurl := fmt.Sprintf("%s%s%d", configJSON.GrafanaBaseURL, configJSON.DeleteAnnotationAPI, annotationID)
+
+	req, err := http.NewRequest("DELETE", apiurl, nil)
+	if err != nil {
+		log.Printf("Request Error, error: %v\n", err)
+		return err
```
vkuznet, line 188: do you really need to return the error here. Can it be replaced with log.Fatal

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+		dump, err := httputil.DumpRequestOut(req, true)
+		if err == nil {
+			log.Println("Request: ", string(dump))
+		}
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		log.Printf("Response Error, error: %v\n", err)
+		return err
```
vkuznet, line 208: will be redundant if you'll replace all return errors with log.Fatal

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		log.Printf("Http Response Code Error, status code: %d", resp.StatusCode)
+		return errors.New("Http Response Code Error")
```
vkuznet, line 212: redundant if you'll replace with log.fatal

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+
+	defer resp.Body.Close()
+
+	if configJSON.Verbose > 2 {
+		dump, err := httputil.DumpResponse(resp, true)
+		if err == nil {
+			log.Println("Response: ", string(dump))
+		}
+	}
+	return nil
```
vkuznet, line 223: will be redundant if all errors will use log.Fatal

vkuznet (changes requested):
Rahul, please review entire code and adjust all functions to get rid of multiple error handling. I only check couple of functions and all of them inherit the same problem.

src/go/MONIT/annotationManager.go
indrarahul2013: Fixed

src/go/MONIT/annotationManager.go
```diff
@@ -0,0 +1,923 @@
...
+		return err
+	}
+
+	err = json.Unmarshal(byteValue, &data)
+	if err != nil {
+		if configJSON.Verbose > 0 {
+			log.Println(string(byteValue))
+		}
+		log.Fatalf("Unable to parse JSON Data from Grafana Annotation GET API, error: %v\n", err)
+		return err
```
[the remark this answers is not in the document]
indrarahul2013: fixed

src/go/MONIT/annotationManager.go
indrarahul2013: fixed.

src/go/MONIT/annotationManager.go
indrarahul2013: fixed.

src/go/MONIT/annotationManager.go
indrarahul2013: fixed.

src/go/MONIT/annotationManager.go
indrarahul2013: fixed.

indrarahul2013:
@vkuznet Fixed error handling issues.

vkuznet:
Rahul, I merged PR, do you still have access to lxplus? If so, can you run and/or coordinate with Christian the script to clean-up the dashboards?

PR #99, opened 2020-09-04
annotationManager - grafana-limit flag added, token read from file feature added

indrarahul2013:


src/go/MONIT/annotationManager.go
```diff
@@ -850,6 +865,7 @@ func parseTimes(ifCreatingAnnotation bool) []int64 {
 func main() {
 
 	flag.StringVar(&annotation, "annotation", "", "Annotation text")
+	flag.StringVar(&grafanaLimit, "grafana-limit", "100", "Limit for fetching number of annotations at a time.")
```
vkuznet, line 868: the limit is an integer type, once you need to pass it to a string you'll use %d in fmt.Sprinf

src/go/MONIT/annotationManager.go
```diff
@@ -40,6 +40,9 @@ var annotation string
 // annotation ID for deleting single annotation
 var annotationID int
 
+// limit for fetching number of annotations at a time
+var grafanaLimit string
```
vkuznet, line 44: the limit is an integer data-type, please do not mix data types when it is not necessary

src/go/MONIT/annotationManager.go
```diff
@@ -117,6 +120,8 @@ func getAnnotations(data *[]annotationData, tags []string) {
 		v.Add("tags", strings.Trim(tag, " "))
 	}
 
+	v.Add("limit", strings.Trim(grafanaLimit, " "))
```
vkuznet, line 123: there is no need for Trim once you convert grafanaLimit to int, just use fmt.Sprintf("%d", grafanaLimit)

vkuznet (changes requested):

src/go/MONIT/annotationManager.go
indrarahul2013: fixed.

src/go/MONIT/annotationManager.go
indrarahul2013: fixed.

src/go/MONIT/annotationManager.go
indrarahul2013: fixed.

PR #100, opened 2020-09-09
monit.go - minor fixes in parseTimes(), addAnnotation(), findDashboard()

indrarahul2013:
- parseTimes()
  - error at day offset, it should be 60\*60\*24
- addAnnotation()
  - response body closed, added error handling
- findDashboard()
  - the tags were just getting set, but it should be added so that all of the tags can get append in the URL.


indrarahul2013:
@vkuznet You can merge this PR.

PR #102, opened 2020-10-20
Add go monit tool query examples

mrceyhun:


doc/MONIT/examples.md
```diff
@@ -0,0 +1,63 @@
+# MONIT go client query examples
+
+## 1. How to use
+1. You can use monit tool directly from `/cvmfs/cms.cern.ch/cmsmon/monit --help`
+2. You can build your own monit go tool.
+    - Firstly download requirements: 
+    
+        `go get github.com/go-stomp/stomp && go get github.com/vkuznet/x509proxy`
```
vkuznet, line 8: before that user(s) required to setup GOPATH, otherwise it will not work. Please add appropriate notice here.

vkuznet (changes requested):

doc/MONIT/examples.md
mrceyhun: I added a line for how to install golang and how to set go path.

doc/MONIT/examples.md
```diff
@@ -0,0 +1,66 @@
+# MONIT go client query examples
+
+## 1. How to use
+1. You can use monit tool directly from `/cvmfs/cms.cern.ch/cmsmon/monit --help`
+2. You can build your own monit go tool.
+    - Firstly, if it is not already exist, [download golang](https://golang.org/doc/install) and set golang path. For linux after downloading: `export PATH=$PATH:/usr/local/go/bin`
```
vkuznet, line 6: this is misleading since golang can be installed in ANY path and user need to setup GOPATH in addition. Instead I suggest to write as following:

Download golang tar ball (for your platofmr) from golang.org, unpack and setup PATH and GOPATH. The PATH should include path to `go` executable, e.g. /usr/local/bin or whatever path you'll choose on your system, and GOPATH can be **any** area on your node which will contain all go dependencies installed in your user space. For instance, you may create you gopath in your home area `mkdir -p $HOME/gopath` and then setup GOPATH points to it, e.g. `export GOPATH=$HOME/gopath`

vkuznet (changes requested):

doc/MONIT/examples.md
mrceyhun: Thank you Valentin. I added.

PR #104, opened 2020-11-19
Edit monit.go such that do not require admin token for queries

mrceyhun:
Two solution came to my mind:
1-) [Implemented in this PR] We can keep raw datasources.json in public place of afs or eos. So simple cronjob can work to update datasources.json:
```
curl -k --location --request GET 'https://monit-grafana.cern.ch/api/datasources' \
--header 'Content-Type: application/json' \
--header "Authorization: Bearer $GRAFANA_ADMIN_TOKEN" > /afs/cern.ch/user/c/cuzunogl/public/grafana/datasources.json
```

2-) We can use https://github.com/dmwm/CMSMonitoring/blob/master/static/datasources.json . However, I could not create json struct in go with this json format. I tried json-to-go and jsonutils libraries also, but did not find a solution either. If we use this kind of json structure, we can fetch necessary datasources information from given github path which we keep updated it:
```
[
    {
        "datasource": "ES_cms_rucio_raw_events",
        "type": "elasticsearch", 
        "id": 9269, 
        "database": "[monit_prod_cms_rucio_raw_events*]"
    }, 
    {
        "datasource": "monit_idb_monitoring",
        "type": "influxdb", 
        "id": 9109, 
        "database": "monit_production_monitoring"
    }
    ...
]
```

### DIFF
--- a/src/go/MONIT/monit.go
+++ b/src/go/MONIT/monit.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -27,6 +28,9 @@ import (
 
 // Record represents MONIT return record {"response":...}
 type Record map[string]interface{}
+
+// DSRecordMaps represents data sources maps {ds1:{id: 1, ...}, ds2:{id:2, ...}}
+type DSRecordMaps map[string]map[string]interface{}
 
 // DSRecord represents record we write out
 type DSRecord struct {
@@ -69,7 +73,7 @@ func read(r string) string {
 }
 
 // return CMS Monitoring datasources
-func datasources(rurl, t string, verbose int) []DSRecord {
+func datasources(rurl, t string, verbose int) ([]DSRecord, error) {
 	uri := fmt.Sprintf("%s/api/datasources", rurl)
 	req, err := http.NewRequest("GET", uri, nil)
 	token := read(t)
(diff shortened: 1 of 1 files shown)


vkuznet:
@mrceyhun , please do not remove code which already produce datasources. Instead of remembering curl command I can easily run monit tool to create datasource file, e.g.
```
./monit -datasources -token ~/certificates/admin.token
[
	{
		"id": 9668,
		"name": "__test_es_condor",
		"type": "elasticsearch",
		"database": "[monit_prod_condor_raw_metric*]"
	},
....
```
It is much easier to do this instead of curl, and if necessary we can create datasource json file just using the monit tool. Yes, it requires admin token, but curl as well. So please close and discard this PR. What we need is simple code which can use `func datasource` output and walk through it to create datasource json format represented in static area. Since output of `func datasources` is list of dict record it is very trivial to convert it to static format, e.g.
```
out := make(map[string]map[string]string
for _, rec := datasources() {
   r := make(map[string]string)
   r["type"] = rec.Type
   r["id"] = rec.Id
   r["database"] = rec.Database
   out[rec.Name] = r
}
```
That's it. So, you can provide another function which convert `[]DSRecord` format into dict one (shown in my code snippet). Then we can use new function to write datasource.json in static area.

mrceyhun:
Thanks a lot for the help @vkuznet 

So, I edited this PR's commit. There are 2 new functions: readDS, writeDS.

writeDS : fetches data sources from Grafana(originals) and then write to a file(if provided) in a way you suggested.

readDS:  
- initializes DataSources object in main function. 
- fetches data sources from [datasources.json](https://github.com/dmwm/CMSMonitoring/blob/master/static/datasources.json) which do not require admin token. I think we do not need extra static file to store data sources, because we are updating datasources.json regularly.

src/go/MONIT/monit.go
```diff
@@ -799,14 +879,21 @@ func main() {
 		log.SetFlags(log.LstdFlags)
 	}
 	var e error
-	DataSources = datasources(defaultUrl, token, verbose)
```
vkuznet, line 802: @mrceyhun it is not correct logic here, we should support both reading datasource with admin token and bypass reading from github and reading from github repo. I suggest the following changes:
- make datasources function to return both datasources and error
- if error is nil then you don't need to readDS, otherwise you will use readDS
Doing this way will accomplish all use-cases:
- users will be able to read from github url
- admins can by pass github url and read datasoruce with admin token.

vkuznet (changes requested):
Everything looks fine to me, but I request additional change for datasources call.

mrceyhun:
@vkuznet 
Admin can read datasources directly from Grafana and users can read it from Github repo.

PR #108, opened 2020-12-08
Added dashboardId filter provided by Grafana annotations API in annotationManager.go tool

indrarahul2013:
Added dashboardId filter provided by Grafana annotations API in annotationManager.go tool
@vkuznet

### DIFF
--- a/src/go/MONIT/annotationManager.go
+++ b/src/go/MONIT/annotationManager.go
@@ -61,6 +61,9 @@ var configFilePath string
 // config variable
 var configJSON config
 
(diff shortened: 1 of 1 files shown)


mrceyhun:
Hi @indrarahul2013 
Thanks a lot! 

There is a small variable formatting problem I guess. I got error:
# command-line-arguments
./annotationManager.go:123:9: cannot use dashboardID (type int) as type string in argument to v.Add

Can you fix this small thing.

indrarahul2013:
@mrceyhun I have fixed the issue. Thanks for pointing out.

PR #116, opened 2021-06-28
Add option to fetch detailed descriptions and comments in the ggus parser

PanosIs:
Add provisions to fetch and parse the full description and comments of GGUS tickets in the parser, option enabled via command line flag. Operation is pretty naive, just combing through the HTML response of GGUS since there is no XML or JSON output option for this information. 

PR adds dependency to "golang.org/x/net"

### DIFF
--- a/src/go/MONIT/ggus_parser.go
+++ b/src/go/MONIT/ggus_parser.go
@@ -14,6 +14,7 @@ import (
 	"net/http/httputil"
 	"os"
 	"os/user"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -301,16 +302,22 @@ func ggusRequest(url, accept string) *http.Response {
 	return resp
 }
 
-func retrieveTicketDetails(ticketId int) (string, []TicketComment) {
+func retrieveTicketDetails(ticketId int, descriptionRgx *regexp.Regexp, commentRgx *regexp.Regexp) (string, []TicketComment) {
 	url := fmt.Sprintf("https://ggus.eu/?mode=ticket_info&ticket_id=%d", ticketId)
 	resp := ggusRequest(url, "text/html")
 	fmt.Println(resp)
 
-	doc, _ := html.Parse(resp.Body)
+	doc, err := html.Parse(resp.Body)
+	if err != nil {
+		log.Println(err.Error())
+		return "", []TicketComment{}
+	}
 
 	description := ""
 	var comments []string
 
+	// Recursively walk the html node tree to find nodes matching the description and comment regexp provided to the
+	// function. The function f needs to be predeclared so that it may later be recursively called.
 	var f func(*html.Node, bool, bool)
 	f = func(n *html.Node, isDescription bool, isHistory bool) {
 		if n.Type == html.TextNode && strings.Trim(n.Data, " \n\t\r") != "" && isDescription {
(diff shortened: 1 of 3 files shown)


src/go/MONIT/ggus_parser.go
```diff
@@ -291,25 +298,88 @@ func processResponse(url, format, accept, out string) {
...
 
+	return resp
+}
+
+func retrieveTicketDetails(ticketId int) (string, []TicketComment) {
+	url := fmt.Sprintf("https://ggus.eu/?mode=ticket_info&ticket_id=%d", ticketId)
+	resp := ggusRequest(url, "text/html")
+	fmt.Println(resp)
+
+	doc, _ := html.Parse(resp.Body)
```
vkuznet, line 309: Panos, please don't skip error, if parser will fail it is better to know why. I suggest to get the error, print it back and return. You may also consider returning error to upstream code, but it will require to change function definition.

src/go/MONIT/ggus_parser.go
```diff
@@ -291,25 +298,88 @@ func processResponse(url, format, accept, out string) {
...
+
+	description := ""
+	var comments []string
+
+	var f func(*html.Node, bool, bool)
+	f = func(n *html.Node, isDescription bool, isHistory bool) {
```
vkuznet, line 315: please add proper comment why you used function in place of another function. Reading the code I see it is recursive but it may not be obvious in a first place.

src/go/MONIT/ggus_parser.go
```diff
@@ -291,25 +298,88 @@ func processResponse(url, format, accept, out string) {
...
+		if n.Type == html.TextNode && strings.Trim(n.Data, " \n\t\r") != "" && isDescription {
+			description = description + strings.Trim(n.Data, " \n\t")
+		} else if n.Type == html.TextNode && strings.Trim(n.Data, " \n\t\r") != "" && isHistory {
+			comments = append(comments, strings.Trim(n.Data, " \n\t"))
+		}
+
+		for c := n.FirstChild; c != nil; c = c.NextSibling {
+			if c.Type == html.CommentNode && c.Data == "#######    Description   #######" {
```
vkuznet, line 323: this is fragile, what if Description string will change? The better approach would be to define regexp patterns which it should match

src/go/MONIT/ggus_parser.go
```diff
@@ -291,25 +298,88 @@ func processResponse(url, format, accept, out string) {
...
+				isDescription = true
+			}
+			if c.Type == html.CommentNode && c.Data == "########## History #########" {
```
vkuznet, line 326: similar comment as before with description

vkuznet (changes requested):

src/go/MONIT/ggus_parser.go
PanosIs: Hey, I agree that this is pretty fragile, but processing this HTML a bit tricky. 

To clarify on your suggestion, do you suggest a regext that matches for example "#+ + Description +#+"? Or something else?

vkuznet: do not hardcode stuff, make it configurable, i.e. provide option to read regexp patterns to use and then code will read them, compile and use to match the html content. Doing this way code will allow to match ANY pattern since it will be configurable from command line.

PanosIs:
Valentin, thanks for the feedback, I have implemented the changes you requested. Please notify me if you have any further notes~

Cheers, 
Panos

PR #124, opened 2021-11-22
Improve eos path size html

mrceyhun:
Update eos path size page


mrceyhun:
@leggerf could you review and merge if it's okay

PR #136, opened 2022-03-31
Add sqoop job for rucio and dbs datasets daily stats

mrceyhun:
- This cron job will dump following tables for the need of sending rucio datasets info to ES task:
    - `FILES, DATASETS, DATA_TIERS, PHYSICS_GROUPS, ACQUISITION_ERAS, DATASET_ACCESS_TYPES` tables from DBS 
    -   and `REPLICAS, CONTENTS` tables from Rucio.
- Since Rucio and DBS oracle instances are separate, dumps will run parallel imports from DBS and Rucio.
- Total dumping time is ~25 minutes for ALL above tables. Total size of all above tables is ~20GB (see example `/tmp/cmssqoop/rucio_daily_stats-2022-03-31`)
- Todays dumps will be deleted in the next run (next day).

Ref: CMSMONIT-450

fyi @leggerf @brij01

### DIFF
--- a/sqoop/scripts/rucio_ds_daily_stats.sh
+++ b/sqoop/scripts/rucio_ds_daily_stats.sh
@@ -1,0 +1,151 @@
+#!/bin/bash
+set -e
+# Reference: some features&logics are copied from sqoop_utils.sh and cms-dbs3-full-copy.sh
+
+##H rucio_ds_daily_stats.sh imports required tables for Rucio datasets daily statistics time series data
+##H
+##H Usage: cd /data/sqoop; ./run.sh ./scripts/rucio_ds_daily_stats.sh
+##H
+##H Script actions:
+##H   None
+##H
+##H Requirements:
+##H   - cmsr_cstring [file]
+##H       should be exist in /data/sqoop directory
+##H       K8s secret name: rucio-secrets
+##H       Origin: secrets/sqoop/cmsr_cstring
+##H   - /etc/secrets/rucio [file]
+##H       K8s secret name: rucio-secrets
+##H       Origin: secrets/rucio/rucio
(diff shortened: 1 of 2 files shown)


sqoop/scripts/rucio_ds_daily_stats.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+set -e
+# Reference: some features&logics are copied from sqoop_utils.sh and cms-dbs3-full-copy.sh
```
vkuznet, line 3: @mrceyhun please add proper documentation to the beginning of this script. It should include:
- all options script can use
- all hard-coded areas, like /etc/secrets/rucio and /data/sqoop/cmsr_cstring as well as their meaning and how they should be setup and/or used
- all env variables you rely on, I see some them use absolute path but other relative ones. I don't know if you check that relative ares, like log, actually exists before using it
- on exit it should print location of the logs to speed up debugging process, in fact if we'll run it on k8s it is better to dump log files on stdout such that we can use kubectl log of the pod without logging to the pod.

vkuznet (changes requested):

sqoop/scripts/rucio_ds_daily_stats.sh
mrceyhun: thanks @vkuznet , I added checks and documentation. My initial idea is to run it in sqoop pod as other sqoop import jobs. Its cron is added to `cronjobs.txt`. 

If I successfully send Spark results to MONIT via Stomp and Rucio colleagues find it useful, maybe we can create a K8s service that runs sqoop import and the Spark job sequentially.

vkuznet (approved):

vkuznet:
@mrceyhun feel free to merge when you are ready

PR #139, opened 2022-04-18
Add locks and rses dumps and delete rucio_ds_daily_stats

mrceyhun:
- No need to `sqoop/scripts/rucio_ds_daily_stats.sh` which is moved to [CMSSpark/bin/cron4rucio_datasets_daily_stats.sh](https://github.com/dmwm/CMSSpark/blob/master/bin/cron4rucio_datasets_daily_stats.sh)
- 2 new tables are added: `RSES` and `LOCKS` as a request from rucio folks
- As a result of a couple of tests, num mappers are decreased to 40 and default fetch size is set (1000) by removing 10k.


vkuznet:
+1

mrceyhun:
Thanks Valentin

PR #142, opened 2022-05-31
Add rules_history and join replicas to main script

mrceyhun:
- Remove `rucio_replicas.sh` and add `replicas` dump to `rucio_table_dumps.sh`. Hence, all Rucio table dumps will be in one script.
- Add `rules_history` daily dump as a request from Oscar.


leggerf:
fine for me

PR #146, opened 2022-06-14
Create build-py GH action for CMSMonitoring python module triggered by tag py-*.*.*

mrceyhun:
New GH action to release CMSMonitoring PyPi module, triggered with **`py-*.*.*`**

- `CMSMonitorin/__init__.py` will not include version as hardcoded anymore. It will be replaced with the git tag version `py-[*.*.*]`

@vkuznet nats-py module requires python_version >= '3.7'. It is only required for nats. StompAMQ and other modules do not have this requirement, so it may break applications which use  python<3.7 and imports CMSMonitoring module. That's why I commented out it it from the `requirements.txt`. We may publish a specific release for `nats` usage or create a new module like `cmsmonitoring_nats` in PyPi. Do you have any suggestions?

fyi @leggerf @brij01


vkuznet:
@mrceyhun , the CMS is already dropped support for python2, recently CMSSW and WMCore were ported to use python 3. As such, I suggest that you evaluate the code and drop py2 support too. For that, you may need new version of nats.py. Since, you have other modules, like StompAMQ.py, I suggest that you convert them first to py3 and then we can come back to this PR and enforces py3 usage. How does it sound?

vkuznet:
you should specify in your requirements that version 3.7 or greater is required, and I'm pretty sure dmwm also needs it. @amaltaro could you please confirm which minimal version of py3 is required for WMCore. According to cmsdist [python3.spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/python3.spec#L1) the version required for builds is 3.8.2

amaltaro:
Talking about a minimal version supported by WMCore is complicated, given that there are many differences between python3 middle version changes (3.6 vs 3.7 vs 3.8).

Having said that, WMCore only supports python3. And being more precise, it's fully compatible and tested with python 3.8.2 (which is what we use, from the spec pointed out by Valentin above).

vkuznet (approved):

PR #151, opened 2022-06-27
Implement custom search and detailed dataset page for each rse

mrceyhun:
There are 2 pages to show different features:
- Old page in http://cmsweb-test1-zone-b-brkegglzfmze-node-1.cern.ch:31280 
  - Uses `datasets` collection
  - Shows datasets and calculations in the RSE that it resides.
  - Does not show individual dataset in RSE details
  - Has `column search` functionality (not works in numeric columns)
- New page in http://cmsweb-test1-zone-b-brkegglzfmze-node-1.cern.ch:31280/details
  - Uses `detailed_datasets` collection
  - Shows each dataset in each RSE with lots of other details.
  - Has `custom search` which is more flexible and has nice UI options like drop down menu
  - New drop down menus for country/tier/kind can be implemented.

Genereal comments:
 - I spent much more time in "details" page, but old one also can be improved. 
- gin-gonic has `ShouldBindJSON` function to get request body and validate it which replaces `io.ReadAll` and json unmarshal.
- Controllers arranged

Problems:
- Because of no caching, paging takes a few seconds (which is not good) . Average query time is less than 5sec, but first opening takes some time. Maybe, it is expectable in ~3.5M documents (`detailed_datasets` collection).
- Naming and code structure is not prod ready. Since I'm trying to prepare this service ASAP, I may miss some things and I kept some hardcoded settings intentionally in test.
- In paging, last page is not working. It's just a skip method, but I could not find enough time to debug it in JS side.

To improve MongoDB performance, I created indexes(compound and single) for `detailed_datasets` manually:
```
db.detailed_datasets.createIndex( { "_id": 1, "Type": 1 } )
db.detailed_datasets.createIndex( { "_id": 1, "Dataset": 1 } )
db.detailed_datasets.createIndex( { "_id": 1, "Dataset": 1, "RSE":1 } )
db.detailed_datasets.createIndex( {"Dataset": 1} )
db.detailed_datasets.createIndex( { "Dataset": 1, "RSE":1 } )
```
I'll observe the performance.


vkuznet (changes requested):
@mrceyhun , I don't have so far specific comments about codebase, but I rather suggest that you add benchmarking tests for all Mongo related functions, like GetSortBson, GetXXX. Since you mentioned about performance (which is most likely due to lack of indexes) we need to know the throughput of your implementation. Please add benchmark tests for all GetXXX functions. For specific examples please see https://github.com/dmwm/dbs2go/blob/master/test/bench_test.go

Moreover, you may have a look at DAS MongoDB layer https://github.com/dmwm/das2go/blob/master/mongo/mongo.go, in particular to Get function and compare your implementation with it. If I recall correctly I was able to achieve 20K req/sec throughput in DAS (using indexes).

Once, I'll know benchmarks it would be easier to advise about the codebase. For instance, your functions perform full data fetch, e.g. you get results from Mongo and return lists. But you may consider streaming as I did in dbs2go, see https://github.com/dmwm/das2go/blob/master/mongo/mongo.go, which provides an alternative way to return data (it has advantage of steady RAM usage regardless of number of returned records as it write results directly to `io.Writer`.

mrceyhun:
Thanks a lot @vkuznet , 
- I implemented `cursor.All()` instead of iterative way `.Next()`. I could not implemented benchmark tests yet, but I will.
- I changed initial sorts, counting filtered queries, pages, etc. There are lots of changes which can be seen in the page directly.
- Searches support both fuzzy search and regex (like `.*AOD.*`), also they are case-insensitive .
- Secret environment load is changed
- In main page, when you click to dataset name in the row, it will show detailed table fetched from detailed collection.

I added 2 examples in [docs/examples.md](https://github.com/dmwm/CMSMonitoring/blob/fea60e7aac98d5a2aa5f8317a0d56c1a408b0c7b/src/go/rucio-dataset-mon-go/docs/examples.md) file. It needs further improvements, but it is working fine for now.

New endpoints for the pages to test:
- http://cmsweb-test1-zone-b-brkegglzfmze-node-1.cern.ch:31280/
- http://cmsweb-test1-zone-b-brkegglzfmze-node-1.cern.ch:31280/rse-details

src/go/rucio-dataset-mon-go/controllers/controllers.go
```diff
@@ -0,0 +1,244 @@
...
+		tempReqBody, _ := c.Get(gin.BodyBytesKey)
+		utils.ErrorResponse(c, "Bad request", err, string(tempReqBody.([]byte)))
+	}
+	return ctx, cancel, start, r
+}
+
+func verboseControllerOutLog(start time.Time, name string, req any, data any) {
+	if Verbose > 0 {
+		elapsed := time.Since(start)
+		req, _ := json.Marshal(req)
```
vkuznet, line 52: please do not skip error checking, especially when parsing json data. It is easy to miss a mailformed json data in this case. I suggest you replace `_` with `err` and check it. According to logic of your code, if it exists you can call `log.Fatal` in that case to provide details about the error.

src/go/rucio-dataset-mon-go/controllers/controllers.go
```diff
@@ -0,0 +1,244 @@
...
+		errCustom := c.ShouldBindBodyWith(&rCustom, binding.JSON)
+		if errCustom == nil {
+			datasetsResp = datasets.GetResults(ctx, c, rCustom)
+			c.JSON(http.StatusOK,
+				datasetsResp,
+			)
+			verboseControllerOutLog(start, "GetDatasets", rCustom, datasetsResp)
+			return
+		}
+		tempReqBody, _ := c.Get(gin.BodyBytesKey)
```
vkuznet, line 97: here is again you are missing error checking

src/go/rucio-dataset-mon-go/controllers/controllers.go
```diff
@@ -0,0 +1,244 @@
...
+		errPlain := c.ShouldBindBodyWith(&rPlain, binding.JSON)
+		if errPlain == nil {
+			datasetsResp = datasets.GetResults(ctx, c, rCustom)
+			c.JSON(http.StatusOK,
+				datasetsResp,
+			)
+			verboseControllerOutLog(start, "GetDatasets", rPlain, datasetsResp)
+			return
+		}
+		tempReqBody, _ = c.Get(gin.BodyBytesKey)
```
vkuznet, line 111: here is again missing error checking

src/go/rucio-dataset-mon-go/controllers/controllers.go
```diff
@@ -0,0 +1,244 @@
...
+	return func(c *gin.Context) {
+		log.SetFlags(log.LstdFlags | log.Lshortfile)
+		verboseControllerInitLog(c)
+		start := time.Now()
+		ctx, cancel := context.WithTimeout(context.Background(), mongo.Timeout)
+
+		// Get request json with validation
+		req := models.ShortUrlRequest{}
+		if err := c.ShouldBindBodyWith(&req, binding.JSON); err != nil {
+			tempReqBody, _ := c.Get(gin.BodyBytesKey)
```
vkuznet, line 143: error check is missing

src/go/rucio-dataset-mon-go/detailed_datasets/detailed_datasets.go
```diff
@@ -0,0 +1,112 @@
...
+	return sortBson
+}
+
+// CreateSearchBson creates search query
+func CreateSearchBson(dtRequest models.DataTableCustomRequest) bson.M {
+	findQuery := bson.M{}
+
+	customs := dtRequest.Custom
+	if customs.Dataset != "" {
+		findQuery["Dataset"] = primitive.Regex{Pattern: customs.Dataset, Options: "im"}
```
vkuznet, line 44: I'm not sure about `primitite.Regex` but you should check if you compile regex each time you call the function or not. If the former, you should avoid that and create compiled regex expression only once to avoid wasting time on that on each function call.

src/go/rucio-dataset-mon-go/controllers/controllers.go
```diff
@@ -0,0 +1,244 @@
...
+		hashId := c.Param("id")
+		log.Printf("[INFO] Hash Id: %s", hashId)
+		shortUrlObj := short_url.GetRequestFromShortUrl(ctx, c, hashId)
+
+		dataTimestamp := data_timestamp.GetDataSourceTimestamp(ctx, c)
+		c.HTML(
+			http.StatusOK,
+			"index.html",
+			gin.H{
+				"title":                "Home Page",
```
vkuznet, line 173: I'm not sure where this structure is defined, but it will be nice if you'll put all data structures in one place and use them here in other places where you call `gin.H`.

src/go/rucio-dataset-mon-go/controllers/controllers.go
```diff
@@ -0,0 +1,244 @@
...
+	defer cancel()
+
+	// get source data creation time
+	dataTimestamp := data_timestamp.GetDataSourceTimestamp(ctx, c)
+	c.HTML(
+		http.StatusOK,
+		"index.html",
+		gin.H{
+			"title":          "Home Page",
+			"is_short_url":   "false",
```
vkuznet, line 224: why `"false"` is a string data-type instead of normal boolean?

vkuznet (changes requested):
I noticed few things which require attention:
- please do not skip error checking, I provided few examples where it is the case
- the `gin.H` uses explicit data, I rather prefer to see data struct in place since data struct clearly define JSON structure , moreover you may add proper validation, e.g. like you used `"false"`  as string while it should be boolean data-type
- all regular expression is desired to compile before calling them to avoid wasting time in function calls
- it would be nice to complement code with architectural diagram representing data and request flow, e.g. see [dbs2go](https://github.com/dmwm/dbs2go/blob/master/docs/DBSServer.md) architecture. This is very valuable when someone will try to understand how code behaves.

src/go/rucio-dataset-mon-go/controllers/controllers.go
mrceyhun: Sorry Valentin, I did not understand which structure. If you mean `gin.H`, it is default `gin-gonic` structure to inject Controller data to html template.

src/go/rucio-dataset-mon-go/controllers/controllers.go
mrceyhun: I put necessary comment for this. It should be `"false"` string since JavaScript function will behave it accordingly. 
`var isShortUrl = ({{.is_short_url}} === "true");`
I don't know if we can pass Go `false` value to JavaScript boolean type. Using string type is one of the secure ways I guess.

mrceyhun:
Thanks a lot @vkuznet . I handled exceptions and put necessary comments to explain  functions/controller clearly. 

- Now `SearchBuilder` is implemented. This is better than custom inputs. Users can define their own conditions
- One of the hard implementation was to implement short URL. I explain how I implemented in the `short_url/short_url.go`. It gets MD5 hash of request and store this hash and request in MongoDB collection. User shares this MD5 hash and another user can restore the exact page of the main user. There are lots of DataTables and JS tricks but it is working fine. I mean "exact" page which is sort order, page number, search queries, etc.; all of them will be exactly same with the second user.
- Data source timestamp is implemented.

- Latest changes are deployed in http://cmsweb-test1.cern.ch:31280/ , please give it a try to all functionalities.
- Here is an example short url: http://cmsweb-test1.cern.ch:31280/short-url/56a8e4d5c049c946d73707a2bb7ba773 . 
  - First condition is `AVG > 10TB`
  - Second condition is `Dataset , regex, GEN-SIM`
  - Third condition is `Rse Type, regex, DISK`
  - And page is **2**
  - hope it is exactly same :)  


@vkuznet , my commit messages are very bad and does not give much information about what is done. If you want me to squash them, I can. I don't know if this is ready to merge but merging will make my life easier for deployment, godoc, etc.

src/go/rucio-dataset-mon-go/data_timestamp/data_timestamp.go
```diff
@@ -0,0 +1,30 @@
...
+
+	cursor, err := mymongo.GetFindOnlyMatchResults(ctx, collection, bson.M{})
+	if err != nil {
+		utils.ErrorResponse(c, "datasetsDb.Find query failed for DataTimeStamp", err, "")
+	}
+	if err = cursor.All(ctx, &dataTimestampLst); err != nil {
+		utils.ErrorResponse(c, "datasets cursor failed", err, "")
+	}
+	// There should be only 1 document
+	return dataTimestampLst[0]
```
vkuznet, line 29: where is a check that `dataTimestampList` will have 1 document? I think it is worth to check it and return proper error message here.

src/go/rucio-dataset-mon-go/datasets/datasets.go
```diff
@@ -0,0 +1,114 @@
...
+			utils.ErrorResponse(c, "TotalRecCount query failed", err, "")
+		}
+		GFilteredCount = cnt
+		GQuery = query
+		log.Println("[INFO] Filter query comparison: MIS-MATCH")
+	} else {
+		// If search query is still same, count should be same, so return GFilteredCount
+		log.Println("[INFO] Filter query comparison: MATCH")
+	}
+	return GFilteredCount
```
vkuznet, line 86: since `GFilteredCount` is global variable you do not need to return it. Therefore, you should either change return o f this function or use local `cnt` variable and make this function independent from globals.

src/go/rucio-dataset-mon-go/detailed_datasets/detailed_datasets.go
vkuznet: ok

vkuznet (changes requested):
i added additional comments after another round of review.

mrceyhun:
Thanks a lot @vkuznet . I implemented all your suggestions except for `GetResults interface` which requires much much more lines than 2 separate functions.
Additionally, controllers are refactored and they are under one folder to improve the readability.

mrceyhun:
@vkuznet since you're on vacation, I would like to merge this PR. Hope it is okay for you and I would like to get your reviews when you're available. Thanks.

PR #161, opened 2022-08-23
Avoid overwritting logger object in StompAMQ modules

### LINKED ISSUE #160: Do not overwrite the logger object in StompAMQ7
With the new python CMSMonitoring version (at least 0.6.5), I noticed that our logger object in ReqMgr2 has a different format.
Looking into the CMSMonitoring source code, here is the override of the object:
https://github.com/dmwm/CMSMonitoring/blob/master/src/python/CMSMonitoring/StompAMQ7.py#L186

I think the best behaviour would be not to change the logger object, if provided. If a new object needs to be created, then yes, it can set that config and format.

amaltaro:
Fixes #160 

Check if a logger object has been passed over to StompAMQ modules before making any changes to it. If one already exists, use it as is without changing the format of the log records.

@mrceyhun could you please review and release this PR. If everything looks fine, I would need a new tag with these changes in.

### DIFF
--- a/src/python/CMSMonitoring/StompAMQ.py
+++ b/src/python/CMSMonitoring/StompAMQ.py
@@ -69,8 +69,11 @@ class StompyListener(object):
     connection.
     """
     def __init__(self, logger=None):
-        logging.basicConfig(level=logging.debug)
-        self.logger = logger if logger else logging.getLogger('StompyListener')
+        if logger:
+            self.logger = logger
+        else:
+            logging.basicConfig(level=logging.debug)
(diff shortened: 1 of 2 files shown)


mrceyhun:
Thanks for the fix @amaltaro , here is the new release that provides the fix `CMSMonitoring==0.6.7` https://pypi.org/project/CMSMonitoring/0.6.7/

amaltaro:
Thanks for the prompt action and for providing a new tag, @mrceyhun

PR #162, opened 2022-08-29
Increase page length limit

mrceyhun:
As requested by our users, new page length options are added `500, 1000, 10000` for "RUCIO DATASETS MONITORING" in http://cmsweb-test1.cern.ch:31280/ . So, max page length option will be `10000`.


mrceyhun:
Image created and deployed `cmsmon-rucio-mon-goweb:rgo-0.0.2`
- `git tag -a rgo-0.0.2 -m "Increase page length limit to 10000"`
- `git push origin rgo-0.0.2`
- Replace service docker image tag and deploy.

PR #163, opened 2022-08-30
Improve sqoop dumps of DBS

mrceyhun:
This PR suggests a more efficient approach to dump DBS tables in `sqoop/scripts/dbs_global_dumps.sh`

Advantages:
- Stores table dumps in HDFS in compressed CSV format
- Totally compatible with `dmwm/CMSSpark/src/python/CMSSpark/schemas.py`
- Sqoop dump total process time of all tables in  `CMS_DBS3_PROD_GLOBAL_OWNER` schema is: `1h 3m 57s` (current one takes more than 1 day)
- Dumps run in parallel
- Less overhead to Oracle because of compressed dumps
- Tables total dump size is "46.7 G", which is great when compared with the size "290.9 G" of `/project/awg/cms/CMS_DBS3_PROD_GLOBAL`
- Test run results can be seen in `cat /data/sqoop/log/2022-08-30_150901_dbs_global_dumps.sh.stdout` in `sqoop-6d7d89657d-kgj6p` K8s pod.
- Test dump results can be seen in `hadoop fs -du -h /tmp/cmssqoop/dbs_global/2022-08-30`

What do you think? If you agree with the current logic, I would like to suggest to change dump schedule to daily. I have a question here, should we continue to use `current/old/touch/new` folder logic, or dump them daily with `YYYY-MM-DD` folder format for each day? 

In second daily option, we can run a cleaner to delete dumps older than 1 month to not exceed our quota. Also it requires to inform our users who currently use `/project/awg/cms/CMS_DBS3_PROD_GLOBAL/(current|old)` folder. So, they should switch to `YYYY-MM-DD` folder format.

### Some concerns about other dumps:

**Concern-1 about HDFS size**
Below snapshot size in HDFS is reached to ~11TB. Should we clean old data?
```
- sqoop/scripts/phedex-blk-replicas-snapshot.sh
$ hadoop fs -du -h /project/awg/cms/phedex/block-replicas-snapshots/
11.2 T  /project/awg/cms/phedex/block-replicas-snapshots/csv
```

There is no problem about quota:
- `/project/awg` quota is 200T, usage is 47T : `hadoop fs -count -q -h /project/awg`
- `/project/awg/cms` quota is infinite: `hadoop fs -count -q -h /project/awg/cms`

So our quota is 200T max.

**Concern-2 about tables which have no data**
- Tables which have no data in "CMS_DBS3_PROD_GLOBAL_OWNER" schema:
	- MIGRATION_REQUESTS
	- MIGRATION_BLOCKS
	- DATASET_RUNS
	- BRANCH_HASHES
	- ASSOCIATED_FILES

Should we continue to try importing them?

#### Some scripts are moved to sqoop/degraded folder because they are not in use anymore:

- sqoop/scripts/cms-jm.sh
  - Was writing to "/project/awg/cms/job-monitoring/avro-snappy"
  - Reason: commented out in cronjobs.txt
- sqoop/scripts/cms-rucio.sh
  - Reason: no usage. probably so old script that dumps ATLAS Rucio replicas table
- sqoop/scripts/cms-rucio-dump.sh
  - Reason: no usage, probably old script for DBS GLOBAL table dumps

- sqoop/scripts/cmssw-popularity.sh
  - Was writing to: /project/awg/cms/cmssw-popularity/avro-snappy
  - Reason: commented out in cronjobs.txt

- sqoop/degraded/jm-cms-data-pop.sh
  - Was writing to: /project/awg/cms/jm-data-popularity/avro-snappy
  - Reason: commented out in cronjobs.txt

**Bottom line**
If you like the suggested approach, I'll apply same logic to PHYS01/2/3 dumps.

@vkuznet you're the first author and who implemented Sqoop dumps at the beginning. That's why, I would like take your opinions.
I can continue to use `current/old/touch/new` folder structure ( so `current` folder will be maximum 1 day old), or daily folder structure. In any case, I'm planning to dump them daily.


vkuznet:
@mrceyhun , your proposal seems solid to me on all fronts. I'm all in favor of reducing size of dumps, cleaning scripts and re-organization. The changes of this PR looks fine to me. But I have doubts about dump policies:
- the dump, especially of very large tables may take long time, in a past if I recall correctly it was taking up to 20 hours
- therefore, I would not go with daily dump until you get enough statistics about dump times.
- which brings another suggestion, please make and collect dump times somewhere, better put it to some exporter and exports to Prometheus and to VM to get visualized. I think you can now easily create JSON data about the dump (per table) and put it on local file and then run json exporter (or similar) to expose to prometheus. You'll figure it out :)

Said that, the MIGRATION_XXX tables are not necessary to dump. I think if you find tables with empty data is most likely not in use in that DBS instance and therefore no need for a dump (but some DBS instances may have use them, e.g. MIGRATION_XXX are heavily used in phys03 DBS but not in global, but anyway they are not required to be dumped).

You may also start monitoring space usage on HDFS and add appropriate alerts. Disk space usage is important for management, therefore I think you can put on your TODO list this task which is not necessary is required here.

Bottom line, please go ahead with this PR, it is ok from my side.

brij01 (commented):
Thanks for this PR @mrceyhun, below are my comments and followup questions. 

>     * Sqoop dump total process time of all tables in  `CMS_DBS3_PROD_GLOBAL_OWNER` schema is: `1h 3m 57s` (current one takes more than 1 day)
Woa.. I didn't knew it took so much time, Thanks for pointing this out and improvement. 

Just for my understanding, I'm wondering about the advantages of sqoop table over mysqldump, is it somehow faster  ?

> What do you think? If you agree with the current logic, I would like to suggest to change dump schedule to daily. I have a question here, should we continue to use `current/old/touch/new` folder logic, or dump them daily with `YYYY-MM-DD` folder format for each day?

Isn't it better to do incremental dumps daily and do full dump weekly while only keeping the last two full dumps ? 
and in my opinion, we should keep them in YYYY-MM-DD format.

> **Concern-1 about HDFS size** Below snapshot size in HDFS is reached to ~11TB. Should we clean old data?

I think we only need to keep last two full dumps and delete the old ones unless there's some special reason. 

> **Concern-2 about tables which have no data**
Importing tables with no data should not be a problem as it will only be preserving the schema and not taking any space.. unless if we don't have any reason to keep these tables in the database itself.

vkuznet:
@brij01 , the `mysqldump` is for MySQL(Maria)DB. While here we talk about ORACLE, so it does not work with ORACLE DB. The incremental dumps requires a timestamp, as such it will require to figure out a last row of each table from a previous dump, and it means a full table scan should be done since table rows may be deleted. The FileLumis table has 22B rows, it means that table scan can takes weeks :) So, the dump is the only option we have, and due to DB size it should go to HDFS, and on HDFS we have sqoop to perform such kind of operations.

mrceyhun:
Thanks @vkuznet , I will export metrics using Pushgateway which is the easiest and suitable solution here IMHO. It inspired me for other metrics too. I will come up with full monitoring of dumps (time spent, # of tables dumped, total dump size, etc.).

@brij01 , fully agree with Valentin's statements. 
Additionally, in past, I tried lots of SQL query solutions which were always slower than full dump. I found this approach after lots of trials a couple of months ago but did not have time to implement it. But now it is a requirement because our users started to use these dumps more frequently and they should be daily. And it is somehow related with K8s cron-jobs.
I can give an example here:
`sqoop eval --connect $JDBS_URL --username $username --password $password --query "select count(*) from CMS_DBS3_PROD_GLOBAL_OWNER.FILE_LUMIS"`
This query is just simple count query on `FILE_LUMIS` table. I tried and it took 35 minutes. With the way of dumping in this PR, dump of whole `FILE_LUMIS` table with 220G size is  between 40-50 minutes. Here are the tricks:
- we use compressed dump `-z/--compressed`
- we use 40 mappers `-m 40`. This is the optimum value in current CERN infra which I inferred after trials (minimised the cost function)
- we are not dumping tables sequentially, all dumps run in parallel so total duration is equal to max dump duration. Data existing check query runs sequentially, that's why total duration is +15 minutes. It is required to check if table has data, because sqoop full dump fails if there is no data in the table.

About data storage procedure, sometimes our users may ask about historical data of some tables like files, datasets, block and couple of others too. May be we can keep them longer(1 year) and others(like file_lumis) 1 month. We have enough quota and we'll have alerts for quota.

mrceyhun:
Now Pushgateway is implemented to send sqoop cron statistics to Prometheus. Functions resides in `utils.sh`. We can follow metrics in this dashboard https://monit-grafana.cern.ch/d/EF3MhXZ4z/sqoop-dumps-statistics?orgId=11 . Currently it includes my test runs. Starting from 8am CERN time, I expect to see real results. The HDFS base directory for these test runs will be `/tmp/cmssqoop/dbs_global/`.

brij01:
> @brij01 , the `mysqldump` is for MySQL(Maria)DB. While here we talk about ORACLE, so it does not work with ORACLE DB. 

Thanks @vkuznet , for Oracle equivalent is OracleDataPump and RMAN.  but yea  I agree with all your points here and perhaps this is the best way to go. 

@mrceyhun, Thanks
>About data storage procedure, sometimes our users may ask about historical data of some tables like files, datasets, block and couple of others too. May be we can keep them longer(1 year) and others(like file_lumis) 1 month. We have enough quota and we'll have alerts for quota.

If you can foresee such a scenario wherein users will ask for legacy data then you can tune the duration of keeping the backup files accordingly. 

Please convert this PR to draft if you're still pushing more changes here.

mrceyhun:
Thanks for the suggestions @brij01 , these tools requires `Oracle Big Data SQL license`. We can open a jira ticket to follow these tools since it requires coordination with Oracle DB. We can ask them if we can use these tools or not in a ticket. There are also other solutions like Flume agents. However, as you know, personally I don't have time to try/test/implement these solutions nowadays, so any help is appreciated :)

mrceyhun:
@vkuznet i would like to ask if `ROW_COUNT`s are useful? I added row count job too to the dump script and it is the cause of today's run which lasted ~3 hours: [dashboard](https://monit-grafana.cern.ch/d/EF3MhXZ4z/sqoop-dumps-statistics?orgId=11). All dumps finished in 57 minutes but count query finished in ~3 hours (even they run parallel). If it is useful, I'll find a better way to count rows of tables.

vkuznet:
@mrceyhun , the `ROW_COUNT` is useful information about table grows if we'll look at historical trend. Said that, I doubt it is useful in this particular case though since now we obtain DBS stats from dbs2go service and you already provided dashboard for it. So we have coverage of this use-case.

mrceyhun:
Thanks @vkuznet , 

I'm dropping support for the calculation of it and for creating `ROW_COUNT` folders in HDFS. If anyone interested, much more enriched DBS table/index statistics can be found in [DBS DB Statistics](https://monit-grafana.cern.ch/d/zqnjuHzVz/dbs-db-statistics?orgId=11) dashboard thanks to @vkuznet

brij01:
> Thanks for the suggestions @brij01 , these tools requires `Oracle Big Data SQL license`. We can open a jira ticket to follow these tools since it requires coordination with Oracle DB. We can ask them if we can use these tools or not in a ticket. There are also other solutions like Flume agents. However, as you know, personally I don't have time to try/test/implement these solutions nowadays, so any help is appreciated :)

Thanks @mrceyhun , I will check about the the Oracle tools and license and followup in a separate PR if needed. 
we're all good here from my side.

mrceyhun:
Hi @vkuznet ,
I think now we can evaluate this PR with more than 14 days of sqoop dump statistics [1] . The statistics belongs to the job that dumps full `CMS_DBS3_PROD_GLOBAL_OWNER` schema tables like [cms-dbs3-full-copy.sh](https://github.com/dmwm/CMSMonitoring/blob/master/sqoop/scripts/cms-dbs3-full-copy.sh) does, including `FILE_LUMIS`,  within ~1 hour. Dump results' base hdfs path is `/tmp/cmssqoop/dbs_global/` .

What do you think about applying this dump method to all dbs sqoop jobs; GLOBAL and PHYS01-3?

Additionally, we observed one of the worst case scenario 2.5 weeks ago with 3.5 hours of max duration. It was caused by `FILE_PARENTS` which is an index-organized table. Contrary to `dataset_parents` and `block_parents`, its size is big. Based on my experience, while finding each mappers row range, sqoop spends so much time because this table doesn't have a PK. There might be also some overload on the DB at that time. In any case, this problem is solved by assigning its mapper as 1. We could've used `--autoreset-to-one-mapper` but in our current script, I think we don't need :)

https://monit-grafana.cern.ch/d/EF3MhXZ4z/sqoop-dumps-statistics?orgId=11

vkuznet (approved):
@mrceyhun , I think you captured everything correctly and based on plots the dumps are under control. I don't have any additional comments at this point.

mrceyhun:
Thanks @vkuznet 

First of all, sorry for force pushing to the same commit. I forgot I was in a conflicted state while making new changes, so I had to amend it and force push :)

While implementing new dumps for DBS schemas, I modified other sqoop jobs. I explained most of them in [README.md](https://github.com/dmwm/CMSMonitoring/blob/15fde4389c26ffced8d949bc00b103881f1d07bc/sqoop/README.md). 

My main aim is to monitor everything and make everything testable in our software stack which will make their automatization more easy. Sqoop is no exception, even though it is one of the heaviest K8s service.

In general:
- I applied pushgateway metric sender for all of them
- Output paths are put into [sqoop/configs.json](https://github.com/dmwm/CMSMonitoring/blob/15fde4389c26ffced8d949bc00b103881f1d07bc/sqoop/configs.json) file. When they are hardcoded in each script, it is hard to automatize Sqoop pod tests. With configurable paths, we can define test output directories and can test all sqoop dumps before deploying to prod. Also this json file's path should be provided  in `CMSSQOOP_CONFIGS` environment variable. Details are in README.md.
- I centralized all secrets directory to `/etc/secrets`, so removed `/etc/cmsdb`. I will make necessary adjustments in Kubernetes service.
- Refactored some codes for readability and shell-checks.
- Created new util functions, see in `sqoop/scripts/utils.sh`


My plan for CMS_DBS_PROD_GLOBAL/PHYS01/PHYS2/PHYS3 schema dumps is that, 
in the new scripts (`dbs3_full_*.sh`) output directory will be `/project/awg/cms/dbs/PROD_(GLOBAL|PHYS0[1-3])`. See in the configs.json. The difference with the previous ones is `/dbs` and short names for sub directories. After testing this PR completely, I will merge it and deploy to prod.

mrceyhun:
Everything is working fine. I will do the switch today. Merging...

leggerf (commented):
ok

PR #164, opened 2022-09-13
Make stomp.py less verbose

amaltaro:
Workaround for https://github.com/dmwm/WMCore/issues/11281

However, tomorrow I am going to suggest a PR upstream to the stomp.py project. So I suggest not to merge these changes until we can hear from the stomp.py developers.

### DIFF
--- a/src/python/CMSMonitoring/StompAMQ7.py
+++ b/src/python/CMSMonitoring/StompAMQ7.py
@@ -76,6 +76,11 @@ class StompyListener(stomp.ConnectionListener):
     def __init__(self, logger=None):
         if logger:
             self.logger = logger
+            # INFO level is very verbose, see: https://github.com/jasonrbriggs/stomp.py/issues/400
+            # bump the log level to the next level to reduce verbosity. FIXME: eventually remove it!
+            thisLogLevel = logging.getLogger("stomp.py").getEffectiveLevel()
+            thisLogLevel = thisLogLevel if thisLogLevel == 50 else thisLogLevel + 10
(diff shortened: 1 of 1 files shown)


src/python/CMSMonitoring/StompAMQ7.py
```diff
@@ -76,6 +76,8 @@ class StompyListener(stomp.ConnectionListener):
     def __init__(self, logger=None):
         if logger:
             self.logger = logger
+            # avoid flood of INFO logs from listener and transport
```
mrceyhun, line 79: @amaltaro  How about making this more flexible. Their logging level is not matching with us according to our usage. Their INFO logs are DEBUG for us, their WARNING is INFO for us. May be we can set stomp log level as one level up for any given logger level.

Let's say you want to debug everything in WMCore, but you'll always get WARNING level logs from Stomp for now.

So, my idea is like: 
```
givenLoggerLevel = logging.getLogger().level
logging.getLogger("stomp.py").setLevel(givenLoggerLevel + 10)
```
- if you give logger with INFO(20) level, stomp one will be WARNING(30).
- if you give logger with DEBUG(10) level, stomp one will be INFO(20).

What do you think?

amaltaro: Sounds good. Let me change this code.

mrceyhun (approved):

mrceyhun:
@amaltaro Changes are fine with me, thanks for the pylint fixes :) Here is the new release from this PR `CMSMonitoring==0.6.8`  : https://pypi.org/project/CMSMonitoring/

PR #165, opened 2022-09-22
Implement new eos quotas page with ec and non ec tables

mrceyhun:
- CMSMONIT-480


mrceyhun:
Table is okay, merging...

Issue #166, opened 2022-09-23
CMSMonitoring python package leaving open file descriptors behind


amaltaro:
As reported in this WMCore GH issue/comment:
https://github.com/dmwm/WMCore/issues/11289#issuecomment-1252335852

we have been having problem with the WMAgent component (AgentStatusWatcher) that uses the CMSMonitoring library to upload data to AMQ.

We are now using version `0.6.8`, which includes heartbeat functionality. So it could be that this issue is either on the CMSMonitoring side or in the `stomp.py` dependency (currently using `7.0.0`).

The previous stable WMAgent release was using CMSMonitoring `0.3.4`, and it was working just fine (it didn't have the heartbeat feature though).

Could you please help us investigating what the problem is?

mrceyhun:
@amaltaro sorry I did not see I was mentioned in the WMCore issue. I am looking it immediately.

mrceyhun:
@amaltaro it can be related with https://github.com/dmwm/CMSMonitoring/blob/master/src/python/CMSMonitoring/StompAMQ7.py#L341 which I intentionally disabled the disconnection after "each send" because it was disconnecting immediately and did not send data. At that time, there was no solution for that. 

It seems this issue is solved in [v8.0.0](https://github.com/jasonrbriggs/stomp.py/compare/v7.0.0...v8.0.0#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR12), let me upgrade our StompAMQ to latest v8.0.1 version and test it.

amaltaro:
Thanks for the prompt feedback, @mrceyhun . If you find a way to provide a CMSMonitoring-only fix, that will be easier to push changes upstream to services already deployed in production.

Upgrade to this latest stomp.py version is likely better to go through our/WMCore usual deployment schedule, if possible.

amaltaro:
@mrceyhun Ceyhun, we have been using in WMCore the cmsmonitoring `0.6.9` version without any problems. Shall we close this ticket?

mrceyhun:
@amaltaro Thanks for reminding. We are also using this version in production without any problem. Fixed with https://github.com/dmwm/CMSMonitoring/pull/167

PR #167, opened 2022-09-23
Add disconnect to StompAMQ7

mrceyhun:
Fix for #166 

Testing took my time. stomp.py still has problems in v8+. In v7.0.0, if we use certificates for authentication, stomp.py produces `ssl EOF occurred in violation of protocol` error on final disconnect. In v8+, even worse, connection does not established with the AMQ broker. I could not find a solution for that and it is more of a certificate problem AFAIK from the community. Anyway, this should work for user/pass auth and it will disconnect after each send().

@amaltaro sorry for the trouble, hope this works! I will be around if you apply this patch today.

### DIFF
--- a/src/python/CMSMonitoring/StompAMQ7.py
+++ b/src/python/CMSMonitoring/StompAMQ7.py
@@ -82,7 +82,8 @@ class StompyListener(stomp.ConnectionListener):
             thisLogLevel = thisLogLevel if thisLogLevel == 50 else thisLogLevel + 10
             logging.getLogger("stomp.py").setLevel(thisLogLevel)
         else:
-            logging.basicConfig(format='%(asctime)s- StompAMQ:%(levelname)s-%(message)s', datefmt='%Y-%m-%dT%H:%M:%S.%f%z',
+            logging.basicConfig(format='%(asctime)s- StompAMQ:%(levelname)s-%(message)s',
+                                datefmt='%Y-%m-%dT%H:%M:%S.%f%z',
                                 level=logging.WARNING)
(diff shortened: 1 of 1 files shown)


amaltaro:
Yes, many of these problems take 10x longer to test than to get a fix done :P Thanks for the prompt fix, Ceyhun.

I applied this patch to the RelVal agent - vocms0258 - and here is some before patch; patching line; post patch file descriptors information, just for the record:
```
# open file descriptors before patching it
$ /data/srv/wmagent/current $ ls -la /proc/1906595/fd | wc -l
1808

# applying the CMSMonitoring patch
$ /data/srv/wmagent/current $ curl https://patch-diff.githubusercontent.com/raw/dmwm/CMSMonitoring/pull/167.patch | patch -d sw/slc7_amd64_gcc630/external/py3-cmsmonitoring/0.6.*/lib/python3.8/site-packages/ -p 3
patching file CMSMonitoring/StompAMQ7.py

# after patching
$ /data/srv/wmagent/current $ ls -la /proc/2549116/fd | wc -l
11
```

amaltaro (approved):
@mrceyhun Ceyhun, I think you nailed it!

I just checked the RelVal agent, and the number of file descriptors is now stable (the same as of a couple of hours ago):
```
$ ls -la /proc/2549116/fd | wc -l
11
```

Could you please make a new release with this fix in (and available in PyPi)? Thanks

mrceyhun:
Great @amaltaro , here it is [CMSMonitoring==0.6.9](https://pypi.org/project/CMSMonitoring/0.6.9/)

PR #169, opened 2022-10-06
Prevent annoying warning email from acron

mrceyhun:


leggerf:
@ceyhun will we still be able to receive errors, if proxy creation fails? It seems to me we're losing this with this change

mrceyhun:
Thanks @leggerf for the heads-up. I missed that part, you are right.
I tried both options (`-q`, `--ignorewarn`), they still produced that warning. I think it is special to `cmsmonit` lxplus. Anyway, I fixed via bash script way. And it is working as expected and produce error in fail.

leggerf (approved):

PR #189, opened 2022-12-15
Implement index page and modify spark jobs to use /project/awg/cms sqoop dumps

mrceyhun:


mrceyhun:
Index page is so simple for now. auth-proxy-server goes to cms-monitoring main page in base path "/" if we don't use an index page.

PR #192, opened 2022-12-21
Move datasets search to search bar for ease of use and make prettier [rucio mon]

mrceyhun:
- Datasets search is moved to search bar instead of SearchBuilder conditions
- Examples are added
- GoDoc link is fixed
- Some html/css cosmetic changes
- Next page numbers are fixed
- Short url base url is fixed: `window.location.href + 'short-url/'`. Previous one: `+ '/short-url/'`

**Breaking change:**  
- Previous short urls will not work.


mrceyhun:
There was a caching problem for static files, It is fixed(hopefully for all users). It is already deployed to production with `rgo-0.0.14` tag.

PR #195, opened 2023-02-12
Add detailed and in-tape-disk pages with many additional features

mrceyhun:
Additional to main datasets page which is a summary table, there are 2 more pages:
`All detailed`: All datasets in each RSE. It is bit slow because of the size of the data.
`In Both Tape and Disk Table`: Datasets in both Tape and Disk:
    - It shows datasets in both Tape and Disk RSEs. Dataset can be partial in them but additional fields provide proper differentiation.
    - Shows "fully locked" and "fully replicated in RSE" details. 
    - "fully locked" means all blocks of the dataset in RSE are locked by production account using. It is calculated  by using `dataset_locks` table.
    - "fully replicated" means all files of the dataset exist in the RSE.
    - It has details button(left green plus button) to show RSE details sorted by Type in which Tape RSE is always on top.

General:
- In "Add Conditions", you will see additional features like "multiple options" for Type and Prod Accounts(lock account). 
- "Dataset" search is always on the tap yellowish input bar and supports both regex and fuzzy "case insensitive" search. For instance, `/nanoaod$` shows only "NANOAOD" datasets.

Technical:
- Features are generalized for all pages. 
- Improves MongoDB indexes
- JS/HTML are improved for better visualization


leggerf (approved):
this is a lot of changes to review...but I assumed you tested/will test this carefully

mrceyhun:
I tested and requested to test from friends. It seems fine. I will do couple of small refactoring, but let's merge this working one.

PR #201, opened 2023-03-01
Add sqoop curator/hdfs_cleaner.sh for retention policy cleaning

mrceyhun:
It seems I forgot to open this PR for HDFS deletion cron for retention policy.


leggerf (approved):
just a detail: why 19 months retention policy for rucio? Didn't we say 18 months?

mrceyhun:
> just a detail: why 19 months retention policy for rucio? Didn't we say 18 months?

It is actually either 18 months or 18 months + 30days. Since I use monthly deletion with regex, if I set 18, retention can be 17 months. May be I should change this to daily deletions...

leggerf:
the cronjob runs daily, right? I guess it does not really matter if data are deleted 18 months -1 day for months with 31 days. We can also say 18 months = 30x18 days, if it makes the logic easier. This is why for example MONIT Kibana retention policy is 30 days, and not 1 month ;)

mrceyhun:
Thanks @leggerf , it is changed to "days ago" calculation.

PR #207, opened 2023-03-20
Add cpueff-goweb stepchain detailed pages, cmsrun+jobtype

mrceyhun:
Hi @vkuznet 

We're moving static html pages of StepChain/Condor CPU efficiencies to this web service, similar to the Rucio dataset monitoring GoLang web service.  When you have time, can you please review the used **WMArchive** fields for StepChain CPU efficiency calculations?

All below depends on [WMArchive/docs](https://github.com/dmwm/WMArchive/tree/master/doc). These are the most critical assumptions and calculations that I would like to get your review:
- I assumed: `TotalJobTime ` is wall clock time of job. So, we can say "threads_total_job_time" as `TotalJobTime * nthreads`
- I assumed: `TotalJobCPU` is total core(all CPUs) time. It will be so trivial, but let's assume core-1 lasted 10 and core-2 lasted 20 seconds, so TotalJobCPU is 30 seconds. Important for following assumption.
- Therefore, cpu efficiency can be calculated as `sum(TotalJobCPU) / sum(TotalJobTime * nthreads) * 100`

All aggregation documentation is here: https://github.com/mrceyhun/CMSMonitoring/blob/fb1767d0e75c7d23827eaeff3f66eb16e88e3489/cpueff-goweb/docs/aggregations.md#stepchain-cpu-efficiency-calculations

Current version is already deployed for review and behind SSO: https://cms-monitoring.cern.ch/web/cpueff . I thought using `web` path for these kind of web services makes sense.


vkuznet:
@mrceyhun , even though I developed WMArchive service I'm not the right person to evaluate the data we store to it. I think the best person to review your description should be @amaltaro . Alan, please review the procedure Ceyhun outlined in this PR, I'll only review the codebase.

mrceyhun:
Thanks Valentin. I expect more discussion over it and we need to show documentation and code-base to our users to review. That's why I'll merge and they can find all the details in the correct repo links.

Merging, but please continue to review ...

PR #217, opened 2023-04-04
Adds Contacts column to Openstack accounting

Germano0:
Hello, I added a new `key:value` to the JSON: it is the `contacts` key

### DIFF
--- a/src/html/openstack_accounting/main.html
+++ b/src/html/openstack_accounting/main.html
@@ -180,7 +180,7 @@
                 // UPDATE WHEN SCHEMA CHANGES!
                 "columnDefs": [
                   {
-                    "targets": [0, 14, 15, 16] ,
+                    "targets": [0, 14, 15, 16, 17] ,
                      "visible": true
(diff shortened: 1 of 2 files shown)


src/html/openstack_accounting/main.html
```diff
@@ -180,7 +180,7 @@ <h3> Openstack Quotas And Usage Monitoring </h3>
                 // UPDATE WHEN SCHEMA CHANGES!
                 "columnDefs": [
                   {
-                    "targets": [0, 14, 15, 16] ,
+                    "targets": [0, 14, 15, 16, 17] ,
```
mrceyhun, line 183: @Germano0 can you update the comment `// DEFAULT SHOW ONLY: ....` and add "contacts" column there too for future reference.

Germano0: @mrceyhun done

mrceyhun:
Thanks! I'll ping you when it is deployed.

PR #228, opened 2023-05-25
Move cmsmonitoringbackup Kubernetes cluster

nikodemas:
CMSMONIT-531

We are moving cmsmonitoringbackup job from to Kubernetes cluster. For this a few changes are needed:
* code needs to be transferred to CMSMonitoring repository (therefore GitLab to GitHub)
* it will be using cmsmon-spark docker [image](https://github.com/dmwm/CMSKubernetes/tree/master/docker/cmsmon-spark) from the Kubernetes so the image needs to be updated accordingly
* authentication and alerting in run.sh needs to be changed as well so that it would work from Kubernetes

### DIFF
--- a/grafana-backup/dashboard-exporter.py
+++ b/grafana-backup/dashboard-exporter.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+import os
+import sys
+import json
+import time
+import shutil
+import tarfile
+import requests
+import subprocess
+import click
+
+baseUrl = 'https://monit-grafana.cern.ch/api'
+
+def updatePathEnding(path):
+    if not path.endswith("/"):
+        path += "/"
+    return path
+def getGrafanaAuth(fname):
+    if not os.path.exists(fname):
+        print("File {} does not exist".format(fname))
+        sys.exit(1)
+    with open(fname, 'r+') as keyFile:
+        SECRET_KEY = json.load(keyFile).get('SECRET_KEY')
+    headers = {"Authorization": "Bearer %s" % SECRET_KEY}
+    return headers
+
+
+def createBackUpFiles(dashboard, folderTitle, headers):
+    allJson = []
+
+    dashboardUid = dashboard.get('uid')
+    path = "grafana/dashboards/%s" % folderTitle
+
+    if not os.path.exists(path):
+        try:
+            os.makedirs(os.path.join('grafana', 'dashboards', folderTitle))
+            print("Successfully created the directory %s " % path)
+        except OSError:
+            print("Creation of the directory %s failed" % path)
+    dashboardUrl = '%s/dashboards/uid/%s' % (baseUrl, dashboardUid)
+    print('dashboardUrl', dashboardUrl)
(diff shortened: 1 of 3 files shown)


mrceyhun:
@nikodemas 
- Let's keep triggering alerts via amtool. I'm adding it to docker image.
- For kerberos authentication, I suggest to source our scripts/utils.sh and use [util_kerberos_auth_with_keytab](https://github.com/dmwm/CMSMonitoring/blob/master/scripts/utils.sh#L131) function. I.e.:
```
in run.sh

. ../scripts/utils.sh
util_kerberos_auth_with_keytab $keytab
```
- You can also use our custom log function `util4logi`, etc..

mrceyhun:
Fyi https://github.com/dmwm/CMSKubernetes/pull/1384
Its GH build will come soon, but we need to merge this PR first to be able to test, so waiting kerberos auth implementation.

grafana-backup/run.sh
```diff
@@ -0,0 +1,61 @@
...
+util_kerberos_auth_with_keytab $1
+
+# Change working directory
+cd "$(dirname "$0")" || exit
+
+addr=ceyhun.uzunoglu@cern.ch
+
+# Add cvmfs envs to run amtool
+export VO_CMS_SW_DIR=/cvmfs/cms.cern.ch
+source $VO_CMS_SW_DIR/cmsset_default.sh
```
nikodemas, line 30: What needs to be done with this part? Do we need to add these scripts separately to the repo since we are moving to Kubernetes?

grafana-backup/run.sh
```diff
@@ -0,0 +1,61 @@
...
+##H        run.sh keytab keys/token.json /cms/backups/grafana/ /eos/cms/store/group/offcomp_monit/
+
+# help definition
+if [ "$1" == "-h" ] || [ "$1" == "-help" ] || [ "$1" == "--help" ] || [ "$1" == "help" ] || [ "$1" == "" ]; then
+    grep "^##H" <"$0" | sed -e "s,##H,,g"
+    exit 1
+fi
+
+# Get utils functions
+. "${script_dir}"/../scripts/utils.sh
```
nikodemas, line 18: I saw that on other occasions utils.sh is copied to the same folder, but is it possible to do it this way?

grafana-backup/run.sh
mrceyhun: We don't need cvmfs which host the pre-built executables. We'll build these executables(amtool, etc.) and use them in docker image.

grafana-backup/dashboard-exporter.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
vkuznet, line 2: All tools should have docstring. Said that, could you please try to write this tool in Go. The code can be smaller, more portable and easy to maintain than Python.

vkuznet (changes requested):

nikodemas:
I have created ticket for the conversion to Go (_CMSMONIT-551_) and merging this PR for now.

PR #236, opened 2023-07-28
Monit tool fix

nikodemas:
CMSMONIT-558

statsquery of the monit tool hasn't been working for a long time because the old endpoint couldn't be reached anymore and therefore it needed to be updated. Changes:
- change endpoint
- include long term cluster information as well
- add error handling for unexpected or empty response from the stats query request
- go tidy dependencies
- create struct for Stats parameters
- delete unused parameters (idx and limit)


src/go/MONIT/monit.go
```diff
@@ -576,6 +579,17 @@ func parseStats(data map[string]interface{}, verbose int) []Record {
 	return out
 }
 
+// helper function to deal with stats query
+func getStats(short_term_url string, long_term_url string, t string, short_term_dbid int, long_term_dbid int, database string, q string, esapi string, idx int, limit int, verbose int, stompConfig StompConfig, inject bool, hdfs string) {
```
vkuznet, line 583: You should define a struct with all of these parameters and pass it around, e.g.
```
type StatsParams struct {
   ShortUrl string
   LongUrl string
}
func getStats(params StatsParams) {
    ...
    long_term_data := run(params.LongURL, ...)
}
```
Doing this way your code will be much easier to read, and you can have as many parameters to pass into function as you like, just extend the struct.

src/go/MONIT/monit.go
```diff
@@ -978,13 +992,35 @@ func main() {
...
-		// 9573, the id of monit_prod_cms-es-size_raw_elasticsearch collection
-		dbid = 9573
-		url = fmt.Sprintf("%s/api/datasources/proxy/%d/_stats/store", url, dbid)
+		// 9573, the id of monit_prod_cms-es-size_raw_elasticsearch datasource, and 9582
+		// the id of monit_prod_condor_agg_metric datasource
+		short_term_dbid := 9573
+		long_term_dbid := 9582
+		short_term_url := fmt.Sprintf("%s/api/datasources/proxy/%d/monit_prod*/_stats/store", url, short_term_dbid)
+		long_term_url := fmt.Sprintf("%s/api/datasources/proxy/%d/monit_prod*/_stats/store", url, long_term_dbid)
+		getStats(short_term_url, long_term_url, t, short_term_dbid, long_term_dbid, database, q, esapi, idx, limit, verbose, stompConfig, inject, hdfs)
```
vkuznet, line 1007: Simplify `getStats` function to use StatsParams, e.g.
```
params := StatsParams{
    ShortUrl: short_term_url,
    LongUrl: long_term_url,
....
}
getStats(params)
```

src/go/MONIT/monit.go
```diff
@@ -978,13 +992,35 @@ func main() {
...
+		// obtain HDFS records if requested
+		if hdfs != "" {
+			records := hdfsDump(hdfs, verbose)
+			injectRecords(stompConfig, records, verbose, inject)
+		}
+		if verbose > 1 {
+			log.Println("short_term_url ", short_term_url)
```
vkuznet, line 1014: If you'll add StatsParams structure then you can define its Print() method to do what this block represents, e.g.
```
type StatsParams struct {
...
}
func (s *StatsParams) Print() {
   log.Println("short_term_url ", s.LongUrl)
   ...
}
```

vkuznet (changes requested):
Please simplify the code for better management and readability by using struct to pass around into getStats function and define its functions, like Print() to do a job.

src/go/MONIT/monit.go
```diff
@@ -410,7 +456,7 @@ func queryURL(rurl string, headers [][]string, verbose int) Record {
 }
 
 // helper function to run ES/InfluxDB or MONIT query
-func run(rurl, token string, dbid int, dbname, query, esapi string, idx, limit, verbose int) Record {
+func run(rurl, token string, dbid int, dbname, query, esapi string, verbose int) Record {
```
vkuznet, line 459: how about further improvement. You may pass StatsParams along with query to run function too instead of individual parameters, and this will simply code and make code further.

vkuznet (changes requested):

src/go/MONIT/monit.go
nikodemas: I am not sure if this works, since `run` method is used not only with `StatsParams`, but also in a more general case for other queries (see [line 1076](https://github.com/dmwm/CMSMonitoring/blob/30bd92762e1c7c23faa453b11fbd64dd944e818e/src/go/MONIT/monit.go#L1076)).

vkuznet: And, why you can't change this line:
```
data := run(url, t, dbid, database, q, esapi, verbose)
```
to something like this:
```
args := StatsParams{Token: t, ...}
data := run(args, q)
```
Bottom line that you don't need **all** parameters to initialize in StatsParams since it will be done automatically according to their data-type. You'll only need to initialize those which you'll use in run function with new signature which will use StatsParams and query. In fact, you may expand StatsParams to include other args, e.g. esapi which is not used by getStats but will be used by run function. You can also do this way:
```
// todo: adjust StatsParams to include esapi
args := NewStats(...)
args.ESAPI = esapi
data = run(args, q)
```
or if you want you can do this way:
```
args = StatsParams{Token: t, ...}
// if you define run function to have StatsParams, esapi and query then you can do
data = run(args, esapi, q)
```

nikodemas: Ok, thank you for your advice!

src/go/MONIT/monit.go
```diff
@@ -60,6 +60,87 @@ func (c *StompConfig) String() string {
...
+		Verbose:   verbose,
+		Config:    config,
+		Inject:    inject,
+		HDFS:      hdfs,
+	}
+}
+
+// builder function of params for queries
+func NewParams(url string, token string, q string, dbname string, dbid int, database string, dbtype string, esapi string, hdfs string, verbose int) *Params {
+	return &Params{
```
vkuznet, line 105: For every parameter of the struct please provide proper validation of data-type, see how it is done in dbs2go code https://github.com/dmwm/dbs2go/blob/master/dbs/datasets.go#L284 and read about Go validator in here https://github.com/go-playground/validator

src/go/MONIT/monit.go
```diff
@@ -410,29 +491,29 @@ func queryURL(rurl string, headers [][]string, verbose int) Record {
 }
 
 // helper function to run ES/InfluxDB or MONIT query
-func run(rurl, token string, dbid int, dbname, query, esapi string, idx, limit, verbose int) Record {
+func run(p Params) Record {
```
vkuznet, line 494: In this function you are passing params by value, i.e. the params struct will be copied. If you'll re-use the same params over and over in different function calls or functions it will make unnecessary copies. Instead, you can change parameters of the function to accept a pointer, i.e. `func run(p *Params)` in this case you will create `Params` structure once and pass a pointer to it to the functions which will make program more efficient (in terms of heap/stack allocation). The beauty of Golang (with respect to C++) is that you can use the same syntax while referring attributes of the struct if you pass a struct itself (by value) or pointer to it, in both cases you will use `p.Query` for example

vkuznet (changes requested):

nikodemas:
Hi @vkuznet, thanks for the review.

I have tried to include parameter validation, but now I think that it is not needed - all of the parameters in the `Params` struct are already validated throughout the code and quite a lot of them are optional as both constructors (`NewStatsParams` and `NewParams`) use the same parent struct. Therefore IMO creating validation methods would not be very useful in this case.

Thank you for the comment about the pointer - I have updated the function parameters.

nikodemas:
@vkuznet do you think I could merge this (based on my last answer)?

vkuznet:
Nikodemas, yes please go ahead with the merge, the changes are good to go.

PR #242, opened 2023-10-05
moves amq sender code to core repo

Panos512:


nikodemas:
@leggerf @brij01 Panos was doing some monitoring stuff and noticed that we are repeating some bits more than once ([here](https://github.com/dmwm/CMSSpark/blob/master/src/python/CMSSpark/rucio_datasets_stats.py#L405-L459) and [here](https://github.com/dmwm/CMSSpark/blob/master/src/python/CMSSpark/rucio_ds_summary.py#L337-L387)) and decided to refactor this. Let us know if you think that these functions should be put somewhere else.

leggerf (approved):
I think it makes sense. Tnx @Panos512

PR #266, opened 2024-03-07
adds eos_accounting_cmsvoc.py

Germano0:
Adds the Python backend script that is run by the CMS VOC on acrontab on his personal account on lxplus.
This script populates the JSON which provides the data shown at https://cmsdatapop.web.cern.ch/cmsdatapop/eos-path-size/size.html

### DIFF
--- a/src/python/CMSMonitoring/eos_accounting_cmsvoc.py
+++ b/src/python/CMSMonitoring/eos_accounting_cmsvoc.py
@@ -1,0 +1,292 @@
+# Germano Massullo - germano.massullo@cern.ch
+# Dario Mapelli - dario.mapelli@cern.ch
+
+# This script is running in acrontab in lxplus under the personal account (not 'cmsvoc' service account) of the CMS VOC, because such account has full permissions.
+# A very extensive explanation of this script is available at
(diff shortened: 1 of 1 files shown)


leggerf (commented):
ok, so are we taking ownership of this? Do we have all the rights needed to run it? Where are we going to run this script?

Germano0:
> Do we have all the rights needed to run it? Where are we going to run this script?
```
# This script is running in acrontab in lxplus under the personal account
(not 'cmsvoc' service account) of the CMS VOC, because such account
has full permissions.
```

leggerf:
yes, I understand this is how this runs now. I am asking how it will run in the future, since I understand Germano is leaving. Is the idea to pass it over to the next VOC, and he/she will run it from his personal account?

I am asking this because this script is filling one of our data sources, but we typically do not own these codes (apart from HTCondor jobs and data management, which are we however trying to pass over to DM group). This part of the data flow is normally responsibility of the team that injects the data (in this case VOC), because they have the expertise to run it and maintain it.

Germano0:
>  Is the idea to pass it over to the next VOC, and he/she will run it from his personal account?

Yes. The reason to save such scripts here is because:

1. archival purposes;
2. the scripts are the backend of the frontend part that is already in this repository, so for consistency purposes I think they should be saved here

leggerf:
great, thanks for the clarfication

PR #277, opened 2024-07-11
Update cms-tasks.sh

nikodemas:
TASKS sqoop dump has been failing because the column `PANDA_RESUBMITTED_JOBS` is not there anymore, so this should be enough to fix it.

FYI @leggerf @brij01 @novicecpp


leggerf (approved):
was a weird name indeed...

PR #280, opened 2024-09-16
Update datasources json

nikodemas:
Update data sources list after the change of OpenSearch aliases


nikodemas:
@novicecpp

novicecpp:
Thanks!

PR #299, opened 2025-07-09
Upgrade jsonschema package as it is not compatible with one used by rucio

vkuznet:
We found incompatibilities of `jsonschema` python package used by rucio and CMS Monitoring. The rucio requires version `jsonschema>=4.20.0` while here we still used `jsonschema~=3.2.0`. Because of this conflict the WM stack cannot be upgraded as it depends both on CMSMonitoring and rucio clients. To align the `jsonschema` version I provide here the upgrade of its dependencies to align with rucio clients.

FYI: @amaltaro

### DIFF
--- a/src/python/requirements.txt
+++ b/src/python/requirements.txt
@@ -1,7 +1,7 @@
 setuptools
 click
 stomp.py>=6.1.1,<8.0.0
-jsonschema~=3.2.0
+jsonschema>=4.20.0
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
Thank you for providing this fix, @vkuznet . 
I think this would solve the WMCore dependency issue (plus, WMCore does not use the schema validation feature in CMSMonitoring, so I am less concerned), so I approve these changes.

However, there has been tons of changes between these major releases, including support to other Python versions, so it is hard to project how it will go.

PR #303, opened 2025-08-28
Add requests table to RUCIO_TABLES variable

Carlosbogo:
As per request of @eachristgr, we are adding the requests table to the Rucio table dumps.

The order of the elements inside the RUCIO_TABLES variable  is important, as they will get snapshotted in that order.
In this case, we want to dump the requests and requests_history tables as close together as possible to avoid any possible data mismatches between them.


eachristgr (approved):
Looks good to me!
Thanks!

PR #309, opened 2026-03-16
Add example code for some Opentelemetry implementations

Carlosbogo:
We are writing some toy examples to be used as templates for other teams implementing their Opentelemetry setup for application level monitoring.

### DIFF
--- a/examples/go_app_otel_example.go
+++ b/examples/go_app_otel_example.go
@@ -1,0 +1,191 @@
+/*
+Toy OpenTelemetry example for a Go application.
+
+What this shows:
+- Traces: creates one span per HTTP request.
+- Metrics: increments a request counter per endpoint.
+- Logs: emits OTLP logs to the collector.
+
+Prerequisites:
+1) Go 1.22+
+2) From this folder, initialize and fetch dependencies:
+    go mod init examples/go-otel-toy
+    go get go.opentelemetry.io/otel@latest
+    go get go.opentelemetry.io/otel/sdk@latest
+    go get go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@latest
+    go get go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc@latest
+    go get go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc@latest
+    go run go_app_otel_example.go
+
+Collector configuration:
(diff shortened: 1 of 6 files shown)


examples/go_app_otel_example.go
```diff
@@ -0,0 +1,191 @@
...
+What this shows:
+- Traces: creates one span per HTTP request.
+- Metrics: increments a request counter per endpoint.
+- Logs: emits OTLP logs to the collector.
+
+Prerequisites:
+1) Go 1.22+
+2) From this folder, initialize and fetch dependencies:
+    go mod init examples/go-otel-toy
+    go get go.opentelemetry.io/otel@latest
```
vkuznet: This is not required if you use:
```
go mod tidy
```
I rather prefer to teach people more modern way to get dependencies instead of show how to explicitly fetch them. I suggest that you test and adjust documentation accordingly.

vkuznet (commented):
This is great, I really like it. I only made comment about go dependencies resolution. It is minor (but in my view) important change.

examples/go_app_otel_example.go
Carlosbogo: Hi Valentin, thank you very much for the advice.
I did not know it could work like that, I haven't used Go in quite a while.

I have updated the code with the changes you requested, let me know if there is anything else you think should be updated.

vkuznet (commented):
I would recommend to do two more things:
- provide README.md describing what is OpenTelemetry, and layout of provided examples (an HTTP server with intrumented telemetry using different Python or GO framework, and a client which can use these metrics)
- and provide image shapshots of how visually see captured metrics

Also, if you are ambition and have spare time, you can provide a docker file(s) for each server and instructions how to deploy it in local k8s cluster where some OTEL server will collect metrics. And, you can provide description and server how to visualize these metrics and use them for your use-case.

Carlosbogo:
Hi again, Valentin, thanks for the reply.

I will add the README.md now to give some more context, and provide some snapshots of Grafana. In terms of documenting the Grafana setup it might be a bit of a more complicated process, because the setup itself is a bit convoluted.

About the docker image, that would be a good idea, but I would not say that the collector is yet in a production state. This makes me a bit hesitant to add it right now (I think it's a great idea for the future), because there is still some work to be done to ensure that these services produce metrics that are actually collected.

I will look into it and see if it's possible to add that.

